PR #1, opened 2020-07-30
Add debug flag

bradleymcallister97:
Add debug flag. Default value is `false`. Can set debug to `true` with `-d` or `--debug`.


internal/clients/flag/flag.go
```diff
@@ -0,0 +1,19 @@
...
+
+import "github.com/spf13/pflag"
+
+// Values contains all values parsed from command line flags
+type Values struct {
+	Debug bool
+}
+
+// Parse parses flags from command line
+func Parse() *Values {
```
bradleymcallister97, line 17: Created a new client for parsing the flags, instead of writing a func in the main file just to keep main from growing very large. Let me know if you feel differently.

internal/clients/diffreviewer/github/github_client.go
```diff
@@ -15,13 +13,6 @@ type Client struct {
 	*github.Client
 }
 
-// NewClient generates a new github client
-func NewClient(httpClientInterface interfaces.HTTPClient) *Client {
```
bradleymcallister97, line 19: Removing unused func

internal/clients/diffreviewer/github/github_service.go
```diff
@@ -124,19 +124,11 @@ type Service struct {
 	CheckRequest *CheckRequest
 }
 
-// NewGithubService creates a new github service with the given httpClient
-func NewGithubService(httpClientInterface interfaces.HTTPClient) diffreviewer.DiffReviewer {
```
bradleymcallister97, line 128: Removing unused func

cmd/nightfalldlp/main.go
```diff
@@ -28,7 +29,9 @@ func main() {
 
 func run() error {
 	ctx := context.Background()
-	diffReviewClient, err := CreateDiffReviewerClient(&http.Client{})
+	flageValues := flag.Parse()
```
alan20854: nit typo in variable name

cmd/nightfalldlp/main.go
```diff
@@ -28,7 +29,9 @@ func main() {
...
+
```
alan20854, line 30: So it looks like we instantiate `flagValues` by calling `flag.Parse` which sets the debug flag to `false`, but how can we set debug to `true`, or will we never need to?

bradleymcallister97: In the flag client we indicate that there is a flag called `debug` or `-d` [here](https://github.com/nightfallai/jenkins_test/pull/1/files#diff-5e93389ce0bcba7cc920f68b25888b55R14). And then when we parse the flags [here](https://github.com/nightfallai/jenkins_test/pull/1/files#diff-5e93389ce0bcba7cc920f68b25888b55R16) the `pflag` lib will set the `Debug` value on the `Values` struct to the appropriate value.

alan20854 (approved):

internal/clients/diffreviewer/github/github_service.go
henkejosh: can we kill the httpClientInterface stuff too? I don't think we were using it anywhere else but not 100% sure

internal/clients/flag/flag.go
```diff
@@ -0,0 +1,19 @@
...
+	values := Values{}
+
+	pflag.BoolVarP(&values.Debug, "debug", "d", false, "Enable debug logs")
```
henkejosh: I assume this func is doing something along the lines of looking for `debug` or `d` CLI arguments and adding whether it was found or not (plus any value following it) to the values struct? If so, I think it probably makes sense to const all these strings at the top of the file for now so that it's super clear what each of them does and so that they're easy to change as we extend the flag functionality.

internal/clients/nightfall/nightfall_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/nightfallai/jenkins_test/internal/clients/diffreviewer"
 	"github.com/nightfallai/jenkins_test/internal/nightfallconfig"
 	"github.com/stretchr/testify/assert"
-	nightfallAPI "github.com/watchtowerai/nightfall_api/generated"
+	nightfallAPI "github.com/nightfallai/nightfall_go_client/generated"
```
henkejosh, line 9: ❤️

internal/nightfallconfig/nightfall_config_test.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"github.com/nightfallai/jenkins_test/internal/nightfallconfig"
 	"github.com/stretchr/testify/assert"
-	nightfallAPI "github.com/watchtowerai/nightfall_api/generated"
+	nightfallAPI "github.com/nightfallai/nightfall_go_client/generated"
```
henkejosh, line 10: ❤️

internal/clients/logger/github_logger/github_logger.go
```diff
@@ -15,19 +16,21 @@ const (
 
 // GithubLogger logger for Github Actions
 type GithubLogger struct {
-	log *log.Logger
+	log   *log.Logger
+	debug bool
```
henkejosh: why does the logger care about this flag - is it so we can only write debug logs if it's true?

bradleymcallister97: I was not exactly sure how the debug flag would be used in Github Actions. I added to the GithubLogger with the thought that if it was true it would print the debug logs. But with Github Actions you still need to set the secret to make debug logs appear.

henkejosh: yeah exactly. my thought is to not add this field to the GithubLogger since we don't explicitly need it, and then we can just use this flag client (and debug flag) for future iterations. What do you think?

henkejosh: we should kill this field right?

henkejosh (approved):

PR #5, opened 2020-08-03
Add help flag

bradleymcallister97:


internal/clients/flag/flag.go
```diff
@@ -13,13 +21,27 @@ type Values struct {
 	Debug bool
 }
 
+func usage() {
+	fmt.Fprint(os.Stderr, "Usage: Nightfall DLP is used to scan content for sensitive information\n\n")
+	pflag.PrintDefaults()
+	os.Exit(2)
```
alan20854: Are exit status codes standardized in any way? Aside from `0` being a successful run

alan20854 (approved):

internal/clients/flag/flag.go
henkejosh: hmm yeah why exit code 2 here? I would think we would want 0 for success

bradleymcallister97: Looking at git and ngrok they return a `0` code for their help command. I am unsure on the effects of returning different error codes. 

This `Usage` function is also called when the `pflag.Parse()` call fails to parse the flags from the command line, or there are unexpected flags.

internal/clients/flag/flag.go
```diff
@@ -14,12 +22,24 @@ type Values struct {
 }
 
 // Parse parses flags from command line
-func Parse() *Values {
+func Parse() (*Values, bool) {
+	fs := pflag.NewFlagSet("all flags", pflag.ContinueOnError)
```
bradleymcallister97, line 26: A flag set needs to be used so that we can set `ErrorHandling = pflag.ContinueOnError` which allows us to catch the parsing error and handle it appropriately.

internal/clients/flag/flag.go
```diff
@@ -14,12 +22,24 @@ type Values struct {
 }
 
 // Parse parses flags from command line
-func Parse() *Values {
+func Parse() (*Values, bool) {
```
henkejosh: can we add a test for this?

bradleymcallister97: Done!

henkejosh (approved):

PR #6, opened 2020-08-03
Make async requests to nightfall API

bradleymcallister97:


internal/clients/nightfall/nightfall_internal_test.go
```diff
@@ -0,0 +1,303 @@
+package nightfall
```
bradleymcallister97, line 1: These tests are not new they have just moved files `nightfall_test.go => nightfall_internal_test.go`.

New tests have been added in `nightfall_test.go`

internal/clients/nightfall/nightfall.go
```diff
@@ -192,6 +196,26 @@ func (n *Client) createScanRequest(items []string) nightfallAPI.ScanRequest {
 	}
 }
 
+func (n *Client) scanContent(ctx context.Context, cts []*contentToScan, reqestNum int, logger logger.Logger) ([]*diffreviewer.Comment, error) {
```
alan20854: nit typo `reqestNum` -> `requestNum`

.github/workflows/nightfalldlp.yml
```diff
@@ -23,13 +23,22 @@ jobs:
           restore-keys: |
```
alan20854: it might be good to just update this entire file to https://github.com/nightfallai/nightfall_dlp_action/blob/445a89d0ea4f1042c290e56c50c3228714d33860/.github/workflows/nightfalldlp.yml

internal/clients/nightfall/nightfall.go
```diff
@@ -227,31 +251,53 @@ func (n *Client) ReviewDiff(ctx context.Context, logger logger.Logger, fileDiffs
...
 
-	comments := []*diffreviewer.Comment{}
-	// Integer round up division
-	numRequestsRequired := (len(contentToScanList) + maxItemsForAPIReq - 1) / maxItemsForAPIReq
-	logger.Debug(fmt.Sprintf("Sending %d requests to Nightfall API", numRequestsRequired))
-	for i := 0; i < numRequestsRequired; i++ {
-		// Use max number of items to determine content to send in request
-		cts := sliceListBySize(i, maxItemsForAPIReq, contentToScanList)
+	commentCh := make(chan []*diffreviewer.Comment)
+	newCtx, cancel := context.WithDeadline(ctx, time.Now().Add(time.Minute*20))
```
henkejosh: probably want to const this timeout duration up top for clarity

internal/clients/nightfall/nightfall.go
```diff
@@ -227,31 +251,53 @@ func (n *Client) ReviewDiff(ctx context.Context, logger logger.Logger, fileDiffs
...
+			blockingCh <- struct{}{}
+			go func(loopCount int, cts []*contentToScan) {
+				defer wg.Done()
+				if newCtx.Err() != nil {
+					return
+				}
+
+				c, err := n.scanContent(newCtx, cts, loopCount+1, logger)
+				if err != nil {
+					cancel()
```
henkejosh: how do we want to handle this case? Should we log for any error (logger.Error) so that it's represented in the program output, but then let us continue on so we can find any additional sensitive info?

internal/clients/nightfall/nightfall.go
```diff
@@ -192,6 +196,26 @@ func (n *Client) createScanRequest(items []string) nightfallAPI.ScanRequest {
...
+	// Pull out content strings for request
+	items := make([]string, len(cts))
+	for i, item := range cts {
+		items[i] = item.Content
+	}
+
+	// send API request
+	resp, err := n.Scan(ctx, logger, items)
+	if err != nil {
+		logger.Debug(fmt.Sprintf("Error sending request number %d with %d items", requestNum, len(items)))
```
henkejosh, line 230: hmm not sure I agree with this being a debug log - (to me) it seems like it would be a good thing to always surface this error to the customer if we encounter it, so I think we either want to explicitly log an Error here, or return and let the parent log an Error.

What do you think?

internal/clients/nightfall/nightfall.go
```diff
@@ -227,31 +251,53 @@ func (n *Client) ReviewDiff(ctx context.Context, logger logger.Logger, fileDiffs
...
+				if err != nil {
+					cancel()
+				} else {
+					commentCh <- c
+				}
+				<-blockingCh
+			}(i, contentSlice)
 		}
+		wg.Wait()
+		close(commentCh)
```
henkejosh: nit - i know there are currently no early exit points in this code block, but it seems like this call to `close` might be good to defer at the top of this go routine (could result in nasty bugs if later on someone adds to this func and ends up exiting early to handle an error or something)

internal/clients/nightfall/nightfall.go
```diff
@@ -227,31 +251,53 @@ func (n *Client) ReviewDiff(ctx context.Context, logger logger.Logger, fileDiffs
...
+	}()
 
-		// Determine findings from response and create comments
-		createdComments := createCommentsFromScanResp(cts, resp, n.DetectorConfigs)
-		comments = append(comments, createdComments...)
+	comments := []*diffreviewer.Comment{}
```
henkejosh, line 318: might help this long, complex func's readability to wrap this bit into a helper func

.nightfalldlp/config.json
```diff
@@ -2,5 +2,6 @@
     "detectors": {
         "CREDIT_CARD_NUMBER": "POSSIBLE",
         "PHONE_NUMBER": "LIKELY"
-    }
+    },
+    "max_number_concurrent_routines": 5
```
henkejosh: why did we choose 5?

internal/nightfallconfig/nightfall_config.go
```diff
@@ -10,19 +10,23 @@ import (
 	nightfallAPI "github.com/nightfallai/nightfall_go_client/generated"
 )
 
+const defaultMaxNumberRoutines = 30
```
henkejosh, line 13: why 30?

internal/clients/nightfall/nightfall.go
bradleymcallister97: Yea, the idea for this debug is that it would give more insight to where exactly the call failed and since the error is returned it is handled appropriately by the parent caller.

.nightfalldlp/config.json
bradleymcallister97: I was testing with test data which required 15 API requests so I used 5 to ensure that the number of go routines was working correctly. 

I can update this if we want to change it.

internal/nightfallconfig/nightfall_config.go
bradleymcallister97: 30 is just an arbitrary number I chose. When we where talking with Rohan he thought that 25-30 go routines seemed like an appropriate default number.

This can be changed to anything we feel it appropriate.

alan20854 (approved):

.nightfalldlp/config.json
```diff
@@ -3,6 +3,7 @@
         "CREDIT_CARD_NUMBER": "POSSIBLE",
         "PHONE_NUMBER": "LIKELY"
     },
+    "max_number_concurrent_routines": 5,
```
henkejosh: nit - camelcase to maintain consistency w/ the other keys here?

internal/clients/nightfall/nightfall.go
```diff
@@ -211,6 +217,59 @@ func (n *Client) createScanRequest(items []string) nightfallAPI.ScanRequest {
...
+		return nil, err
+	}
+
+	// Determine findings from response and create comments
+	createdComments := createCommentsFromScanResp(cts, resp, n.DetectorConfigs, n.TokenExclusionSet)
+	logger.Debug(fmt.Sprintf("Got %d annotations for request #%d", len(createdComments), requestNum))
+	return createdComments, nil
+}
+
+func (n *Client) scanAllContent(ctx context.Context, logger logger.Logger, cts []*contentToScan, commentCh chan []*diffreviewer.Comment) {
```
henkejosh: nit - i think we can be explicit about this channel being write only (`commentCh chan <- []*diffreviewer.Comment`)

internal/clients/nightfall/nightfall_test.go
```diff
@@ -1,347 +1,163 @@
...
+		Title:      fmt.Sprintf("Detected %s", nightfallAPI.CREDIT_CARD_NUMBER),
 	}
+	expectedComments := []*diffreviewer.Comment{&c, &c, &c}
 
-	for _, tt := range tests {
-		actual := createCommentsFromScanResp(tt.haveContentToScanList, tt.haveScanResponseList, detectorConfigs, tt.haveTokenExclusionSet)
-		assert.Equal(t, tt.want, actual, fmt.Sprintf("Incorrect response from createCommentsFromScanResp: %s test", tt.desc))
+	for i := 0; i < numScanReq; i++ {
+		mockAPIClient.EXPECT().ScanAPI().Return(mockScanAPI)
+		mockScanAPI.EXPECT().ScanPayload(gomock.Any(), gomock.Any()).Return(scanResp, nil, nil)
```
henkejosh: any way we can do stricter type checking for the 2nd val passed to ScanPayload here?

internal/clients/nightfall/nightfall.go
```diff
@@ -211,6 +217,59 @@ func (n *Client) createScanRequest(items []string) nightfallAPI.ScanRequest {
 	}
 }
 
+func (n *Client) scanContent(ctx context.Context, cts []*contentToScan, requestNum int, logger logger.Logger) ([]*diffreviewer.Comment, error) {
```
henkejosh, line 220: we don't have to do this for this PR (so we can get this code in :) ) but do you mind creating a ticket for adding tests for this func later? Seems pretty complex and a good place to add some unit tests when we have time.

henkejosh (approved):

PR #7, opened 2020-08-04
Token Exclusion Rules

alan20854:
https://nightfall.atlassian.net/jira/software/projects/NIGH/boards/19?selectedIssue=NIGH-69

Passing in list of items to whitelist from `config.json`, converting to a map, and not creating a comment if the finding fragment is in the whitelist map.

### DIFF
--- a/.nightfalldlp/config.json
+++ b/.nightfalldlp/config.json
@@ -2,5 +2,9 @@
     "detectors": {
         "CREDIT_CARD_NUMBER": "POSSIBLE",
         "PHONE_NUMBER": "LIKELY"
-    }
+    },
+    "tokenWhitelist": [
+        "4242-4242-4242-4242",
+        "4916-6734-7572-5015"
+    ]
 }
--- a/.nightfalldlp/config.json
+++ b/.nightfalldlp/config.json
@@ -3,7 +3,7 @@
         "CREDIT_CARD_NUMBER": "POSSIBLE",
         "PHONE_NUMBER": "LIKELY"
     },
-    "tokenWhitelist": [
+    "tokenExclusionList": [
         "4242-4242-4242-4242",
         "4916-6734-7572-5015"
     ]
--- a/internal/clients/nightfall/nightfall.go
+++ b/internal/clients/nightfall/nightfall.go
@@ -39,18 +39,21 @@ var likelihoodThresholdMap = map[nightfallAPI.Likelihood]int{
 // Client client which uses Nightfall API
 // to determine findings from input strings
 type Client struct {
-	APIClient       *nightfallAPI.APIClient
-	APIKey          string
-	DetectorConfigs nightfallconfig.DetectorConfig
+	APIClient         *nightfallAPI.APIClient
+	APIKey            string
+	DetectorConfigs   nightfallconfig.DetectorConfig
+	TokenWhitelistMap map[string]bool
 }
 
 // NewClient create Client
 func NewClient(config nightfallconfig.Config) *Client {
 	APIConfig := nightfallAPI.NewConfiguration()
+	tokenWhitelistMap := makeStrMap(config.TokenWhitelist)
 	n := Client{
-		APIClient:       nightfallAPI.NewAPIClient(APIConfig),
-		APIKey:          config.NightfallAPIKey,
-		DetectorConfigs: config.NightfallDetectors,
+		APIClient:         nightfallAPI.NewAPIClient(APIConfig),
+		APIKey:            config.NightfallAPIKey,
+		DetectorConfigs:   config.NightfallDetectors,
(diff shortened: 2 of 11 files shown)


internal/clients/nightfall/nightfall.go
```diff
@@ -62,9 +65,11 @@ type contentToScan struct {
 }
 
 func foundSensitiveData(finding nightfallAPI.ScanResponse, detectorConfigs nightfallconfig.DetectorConfig) bool {
-	minimumLikelihoodForDetector := detectorConfigs[nightfallAPI.Detector(finding.Detector)]
```
alan20854, line 65: Prior implementation was also returning true if the finding detector wasn't in detectorConfigs. 
`likelihoodThresholdMap[minimumLikelihoodForDetector]` would default to zero, causing the return value to always evaluate to true.

bradleymcallister97 (approved):

alan20854:
Also tested this with the `nightfall_dlp_action`, pointing to a branch that downloads the `jenkins_test` from this branch. You can see the check with commit hash `b84484e` had no findings, since the sensitive items it is catching (in previous and later commits) are in the config tokenWhitelist

internal/clients/nightfall/nightfall.go
```diff
@@ -249,9 +272,20 @@ func (n *Client) ReviewDiff(ctx context.Context, logger logger.Logger, fileDiffs
...
 		// Determine findings from response and create comments
-		createdComments := createCommentsFromScanResp(cts, resp, n.DetectorConfigs)
+		createdComments := createCommentsFromScanResp(cts, resp, n.DetectorConfigs, n.TokenWhitelistMap)
 		comments = append(comments, createdComments...)
 	}
 
 	return comments, nil
 }
+
+func makeStrMap(strSlice []string) map[string]bool {
```
alan20854: is it worth just using `strset.Set` instead of creating my own set/map?

henkejosh: yup definitely - less code for us to test/maintain :)

internal/clients/nightfall/nightfall.go
```diff
@@ -154,13 +159,19 @@ func sliceListBySize(index, numItemsForMaxSize int, contentToScanList []*content
...
+	resp [][]nightfallAPI.ScanResponse,
+	detectorConfigs nightfallconfig.DetectorConfig,
+	tokenWhitelist map[string]bool,
+) []*diffreviewer.Comment {
 	comments := []*diffreviewer.Comment{}
 	for j, findingList := range resp {
 		for _, finding := range findingList {
-			if foundSensitiveData(finding, detectorConfigs) {
+			if foundSensitiveData(finding, detectorConfigs) &&
+				!isFindingInTokenWhitelist(finding.Fragment, tokenWhitelist) {
```
henkejosh: just to clarify - the `finding.Fragment` is just the sensitive token, no pre or post-context included, right?

alan20854: that's correct

.nightfalldlp/config.json
```diff
@@ -2,5 +2,9 @@
     "detectors": {
         "CREDIT_CARD_NUMBER": "POSSIBLE",
         "PHONE_NUMBER": "LIKELY"
-    }
+    },
+    "tokenWhitelist": [
```
henkejosh: just confirming - is this the same nomenclature the API team uses to refer to whitelists/exclusion rule stuff?

test/data/nightfall_test_config.json
```diff
@@ -2,5 +2,6 @@
     "detectors": {
         "CREDIT_CARD_NUMBER": "POSSIBLE",
         "PHONE_NUMBER": "LIKELY"
-    }
+    },
+    "tokenExclusionList": ["4242-4242-4242-4242", "xG0Ct4Wsu3OTcJnE1dFLAQfRgL6b8tIv"]
```
bradleymcallister97, line 6: I guess the nightfalldlp annotation is showing up since we are using the action and not the code in this diff

bradleymcallister97:
I think we should be using the original workflow where we build and run this repo for nightfalldlp action, instead of using the action from the other repo. This would allow us to test our diffs. Thoughts?

test/data/nightfall_test_config.json
henkejosh: yup exactly

henkejosh (approved):

alan20854:
> I think we should be using the original workflow where we build and run this repo for nightfalldlp action, instead of using the action from the other repo. This would allow us to test our diffs. Thoughts?

Sure that makes sense - ill change it back.

bradleymcallister97 (approved):

PR #8, opened 2020-08-04
Setup Docker Image

bradleymcallister97:


.github/workflows/nightfalldlp.yml
```diff
@@ -1,49 +1,21 @@
...
-        continue-on-error: true
+      - name: nightfallDLP action step
+        uses: nightfallai/nightfall_dlp_action@v0.0.3
         env:
-          NIGHTFALL_API_KEY: ${{ secrets.NIGHTFALL_API_KEY }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          nightfalldlp
+          NIGHTFALL_API_KEY: ${{ secrets.NIGHTFALL_API_KEY }}
+          EVENT_BEFORE: ${{ github.event.before }}
```
henkejosh: nit can we add a newline?

Dockerfile
```diff
@@ -22,3 +12,15 @@ RUN wget -O- -nv https://raw.githubusercontent.com/golangci/golangci-lint/master
 RUN wget -O - -q https://raw.githubusercontent.com/reviewdog/reviewdog/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v0.9.15
 
 COPY . .
+
+RUN CGO_ENABLED=1 GOOS=linux go build -a -installsuffix cgo -o nightfall_dlp ./cmd/nightfalldlp/
+
+FROM alpine:3.8
```
henkejosh, line 18: hmm how does having this 2nd image in here work?

bradleymcallister97: This is a [multi-stage docker build](https://docs.docker.com/develop/develop-images/multistage-build/#name-your-build-stages). So the first docker image is built, and inside that one we get everything required to run ci jobs and build the go binary (do all the heavy lifting). Then the second image is the image that is the final image that we would push to docker hub, which is tiny as it only contains the binary to run the action.

Using the multi-stage build allows us to generate a small final image to be run. It being small is beneficial as it can be pulled quickly from docker hub and it will take up less memory.

Makefile
```diff
@@ -11,6 +11,11 @@ SERVICE_NAME=nightfall_dlp
 BINARY_NAME=./$(SERVICE_NAME)
 GO_TEST_ENV?=test
 
+NAME=nightfallai/nightfall-dlp
```
henkejosh: should this be an underscore?

bradleymcallister97: I created the docker hub repo with the dash but will update to use underscore, since probably makes more sense

.circleci/config.yml
```diff
@@ -51,6 +51,28 @@ jobs:
...
+  push-to-dockerhub:
+    docker:
+      - image: deliveroo/circleci:0.2.8
+    steps:
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run:
+          name: build
+          command: ci build
```
henkejosh, line 63: why `ci build` instead of `make dockerbuild`?

henkejosh (approved):

alan20854 (approved):

.circleci/config.yml
bradleymcallister97: The `ci` command is a helper command provided by the base image `deliveroo/circleci:0.2.8` https://github.com/deliveroo/circleci#custom-helper-ci. It allows us to use the `docker-compose.ci.yml` file.

cmd/nightfalldlp/entrypoint.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh -l
```
bradleymcallister97, line 1: A downside of the script is that it is currently very Github Actions specific

henkejosh: yup. It's not a big deal now since it's pretty simple, but another downside is the fact that most of the team isn't super fluent in bash. Is it worth making a ticket to migrate this script to go? Or should we wait until it becomes a problem/we need to update this file?

henkejosh (approved):

cmd/nightfalldlp/entrypoint.sh
bradleymcallister97: Not sure how difficult it is to migrate it to go, but I think it is worthwhile. Right now the docker image is Github Action specific so if we move it into the go code then the image can be used across all environments that we support.

Created a ticket for it

PR #11, opened 2020-08-06
NIGH-68 File In/Exclusion Rules

alan20854:
https://nightfall.atlassian.net/jira/software/projects/NIGH/boards/19?selectedIssue=NIGH-68

Filtering files to scan by regex. 
If `fileInclusionList` is not empty, only scan the files that match the regex patterns in that list. 
Then, remove the files that match the patterns in `fileExclusionList` to omit scanning on.


internal/clients/nightfall/nightfall.go
```diff
@@ -278,3 +284,34 @@ func (n *Client) ReviewDiff(
...
+	if fileExcludeList != nil && len(fileExcludeList) > 0 {
+		filteredFileDiffs = filterByFilePath(filteredFileDiffs, fileExcludeList, false)
+	}
+	return filteredFileDiffs
+}
+
+func filterByFilePath(fileDiffs []*diffreviewer.FileDiff, regexPatterns []string, include bool) []*diffreviewer.FileDiff {
+	filteredFileDiffs := []*diffreviewer.FileDiff{}
+	for _, fd := range fileDiffs {
+		if (matchRegex(fd.PathNew, regexPatterns) && include) ||
```
alan20854, line 302: if this is for file inclusion, append to `filteredFileDiffs` if the filename matches a regex pattern `or` if this is for file exclusion, append to `filteredFileDiffs` if the filename doesn't match any of the regex patterns

internal/clients/nightfall/nightfall.go
```diff
@@ -278,3 +284,34 @@ func (n *Client) ReviewDiff(
...
+	}
+	if fileExcludeList != nil && len(fileExcludeList) > 0 {
+		filteredFileDiffs = filterByFilePath(filteredFileDiffs, fileExcludeList, false)
+	}
+	return filteredFileDiffs
+}
+
+func filterByFilePath(fileDiffs []*diffreviewer.FileDiff, regexPatterns []string, include bool) []*diffreviewer.FileDiff {
+	filteredFileDiffs := []*diffreviewer.FileDiff{}
+	for _, fd := range fileDiffs {
```
alan20854, line 359: if this is for file inclusion, append to `filteredFileDiffs` if the filename matches a regex pattern `or` if this is for file exclusion, append to `filteredFileDiffs` if the filename doesn't match any of the regex patterns

.nightfalldlp/config.json
```diff
@@ -6,5 +6,7 @@
     "tokenExclusionList": [
         "4242-4242-4242-4242",
         "4916-6734-7572-5015"
-    ]
+    ],
+    "fileInclusionList": [".*"],
+    "fileExclusionList": ["exclude\\.txt"]
```
bradleymcallister97: Im not sure I understand why there is `exclude\\` in the front of this.

Oh is this a specific regex match?

internal/clients/nightfall/nightfall.go
```diff
@@ -278,3 +284,34 @@ func (n *Client) ReviewDiff(
...
+		if (matchRegex(fd.PathNew, regexPatterns) && include) ||
+			(!matchRegex(fd.PathNew, regexPatterns) && !include) {
```
bradleymcallister97: I feel like we should make the call to `matchRegex` outside the if and store the result to be used in the if statement. Otherwise, I believe that we could potentially be making 2 calls to `matchRegex` with the same params.

internal/clients/nightfall/nightfall_test.go
```diff
@@ -302,6 +302,72 @@ func TestBlurContent(t *testing.T) {
 	}
 }
 
+func TestFilterFileDiffs(t *testing.T) {
+	filePaths := []string{"path/secondary_path/file.txt", "a.go", "a/a.go", "test.go", "path/main.go", "path/test.py"}
+	var fileDiffs []*diffreviewer.FileDiff
```
bradleymcallister97: nit - since we know the size of `filePaths` we can make an array of the same size for `fileDiffs`

.nightfalldlp/config.json
alan20854: ya i made a text file named `exclude.txt`, the `\\` is to treat the `.` as the literal dot

test/data/file_exclusion/exclude.txt
```diff
@@ -0,0 +1 @@
+exclude me my phone number is 301-352-4729
```
bradleymcallister97: Just need to remove this before merging

bradleymcallister97 (approved):

internal/clients/nightfall/nightfall.go
```diff
@@ -278,3 +284,34 @@ func (n *Client) ReviewDiff(
 
 	return comments, nil
 }
+
+func filterFileDiffs(fileDiffs []*diffreviewer.FileDiff, fileIncludeList, fileExcludeList []string) []*diffreviewer.FileDiff {
+	filteredFileDiffs := fileDiffs
```
henkejosh: do we need this extra var or can we just filter the actual `fileDiffs` slice?

internal/clients/nightfall/nightfall.go
```diff
@@ -278,3 +284,34 @@ func (n *Client) ReviewDiff(
...
+		filteredFileDiffs = filterByFilePath(filteredFileDiffs, fileIncludeList, true)
+	}
+	if fileExcludeList != nil && len(fileExcludeList) > 0 {
+		filteredFileDiffs = filterByFilePath(filteredFileDiffs, fileExcludeList, false)
+	}
+	return filteredFileDiffs
+}
+
+func filterByFilePath(fileDiffs []*diffreviewer.FileDiff, regexPatterns []string, include bool) []*diffreviewer.FileDiff {
+	filteredFileDiffs := []*diffreviewer.FileDiff{}
```
henkejosh: let's make this slice (using `make`) and specifying a capacity hint (maybe jus the len of `fileDiffs`) so that we hopefully never have to reallocate space for the array
e.g. `make([]*diffreviewer.FileDiff, 0, len(fileDiffs)`

https://github.com/uber-go/guide/blob/master/style.md#specifying-slice-capacity

internal/clients/nightfall/nightfall.go
```diff
@@ -278,3 +284,34 @@ func (n *Client) ReviewDiff(
...
+	for _, fd := range fileDiffs {
+		matched := matchRegex(fd.PathNew, regexPatterns)
+		if (matched && include) || (!matched && !include) {
```
henkejosh, line 363: can we add a comment explaining what these 2 conditions are? I think the logic is sound just a little confusing at first

internal/clients/nightfall/nightfall.go
```diff
@@ -278,3 +284,34 @@ func (n *Client) ReviewDiff(
...
+			filteredFileDiffs = append(filteredFileDiffs, fd)
+		}
+	}
+	return filteredFileDiffs
+}
+
+func matchRegex(filePath string, regexPatterns []string) bool {
```
henkejosh: can we just add a few unit tests for this method? I know it's kind of covered in the TestFilterFileDiffs but would be good to get explicit tests to make sure this piece is doing exactly what we want it to

test/data/file_exclusion/exclude.txt
henkejosh: yup let's kill this

alan20854:
The github service uses `filepath.Match`, whereas I'm using `regexp.MatchString` instead. I'm not sure `filepath.Match` is the best option for users, but I'm open to other opinions.

The docs for `filepath.Match` specify `Match requires pattern to match all of name, not just a substring.`. In addition, the `*` wildcard `matches any sequence of non-Separator characters`. An example of this limitation would be if you had the filepaths `"path/a.txt", "path/secondary_path/b.txt", "path/secondary_path/tertiary_path/c.txt"`, a pattern of `path/*` would only match the first filepath.

To match all three filepaths, you would need to have the following patterns `"path/*", "path/*/*", and "path/*/*/*"`, which does not seem particularly good to require users to add in the config, rather than just `"path/.*"` with `regexp.MatchString`

bradleymcallister97 (approved):

internal/clients/nightfall/nightfall.go
```diff
@@ -328,3 +334,35 @@ func (n *Client) ReviewDiff(
...
+		// if include (file inclusion), append if the filename matches a regex pattern
+		// if !include (file exclusion), append if the filename does not match any pattern
+		if (matched && include) || (!matched && !include) {
+			filteredFileDiffs = append(filteredFileDiffs, fd)
+		}
+	}
+	return filteredFileDiffs
+}
+
+func matchRegex(filePath string, regexPatterns []string) bool {
```
henkejosh: did you have a chance to look into globbing vs regex? thoughts on doing one vs the other?

This library might be useful? https://golang.org/pkg/path/filepath/#Match

alan20854: Yup, if you look at the comment above, `filepath.Match` doesn't seem as user friendly for specifying subdirs. 

In addition, we probably can't use `filepath.Glob` (which uses the same expression syntax as `Match`) either, since it's looking locally for files that match the patterns, which doesn't translate to our use case where we're essentially filtering through a list of filenames.

henkejosh: sorry which comment are you referring to?

And what are your thoughts on pros/cons of globbing vs. regex for this use case?

alan20854:
> The github service uses `filepath.Match`, whereas I'm using `regexp.MatchString` instead. I'm not sure `filepath.Match` is the best option for users, but I'm open to other opinions.
> 
> The docs for `filepath.Match` specify `Match requires pattern to match all of name, not just a substring.`. In addition, the `*` wildcard `matches any sequence of non-Separator characters`. An example of this limitation would be if you had the filepaths `"path/a.txt", "path/secondary_path/b.txt", "path/secondary_path/tertiary_path/c.txt"`, a pattern of `path/*` would only match the first filepath.
> 
> To match all three filepaths, you would need to have the following patterns `"path/*", "path/*/*", and "path/*/*/*"`, which does not seem particularly good to require users to add in the config, rather than just `"path/.*"` with `regexp.MatchString`

@henkejosh this one

internal/clients/nightfall/nightfall.go
alan20854: oops it's on the `Conversation` tab

alan20854: But the main comparison is that globbing has better performance, and regex has easier usability (for users to specify in the config).

test/data/nightfall_test_config.json
```diff
@@ -5,5 +5,6 @@
         "IP_ADDRESS": "POSSIBLE"
     },
     "maxNumberConcurrentRoutines": 20,
-    "tokenExclusionList": ["4242-4242-4242-[0-9]{4}", "xG0Ct4Wsu3OTcJnE1dFLAQfRgL6b8tIv", "^127\\."]
+    "tokenExclusionList": ["4242-4242-4242-[0-9]{4}", "xG0Ct4Wsu3OTcJnE1dFLAQfRgL6b8tIv", "^127\\."],
```
alan20854, line 8: comment: make sure to update token matching to use `matchRegex`

bradleymcallister97 (approved):

README.md
```diff
@@ -9,3 +9,23 @@ security, and ensure you never accidentally leak secrets or other sensitive info
...
+### File Inclusion/Exclusion
+To omit files from being scanned, you can add the `fileExclusionList` field to your config. In addition, to only scan on certain files, add the `fileInclusionList` to the config.
+
+Here's an example use case:
+```
+    fileExclusionList: ["*/tests/*"],
+    fileInclusionList: ["*.go", "*.json"]
+```
+In the example, we are ignoring all file paths with a `tests` subdirectory, and only scanning on `go` and `json` files.
+Note: we are using [gobwas/glob](https://github.com/gobwas/glob) to match file path patterns. Unlike the token regex matching, file paths must be completely matched by the given pattern. e.g. If `tests` is a subdirectory, it will not be matched by `test/*`, which is only a partial match.
```
henkejosh: nit - should this say "it will not be matched by `tests/*`"

henkejosh (approved):

PR #12, opened 2020-08-06
NIGH-73 Regex Token Exclusion

alan20854:
Updating how we're doing token exclusion from hard string matching to regex


alan20854:
more tests to be added

alan20854:
note: need to rebase off brad's changes

internal/clients/nightfall/nightfall_internal_test.go
```diff
@@ -227,15 +231,41 @@ func TestCreateCommentsFromScanResp(t *testing.T) {
...
 				},
 			},
-			haveTokenExclusionSet: tokenExclusionSet,
+			haveTokenExclusionList: tokenExclusionList,
 			want: []*diffreviewer.Comment{
 				createComment(creditCardResponse),
 			},
 			desc: "single credit card excluded",
 		},
+		{
```
alan20854, line 240: added test where exclusion list is `["4242-4242-4242-[0-9]{4}", "^127\\."]`, and detectors are `CREDIT_CARD` and `IP_ADDRESS`

bradleymcallister97 (approved):
Nice!

I'm not exactly sure where this would go, but I think it might be worthwhile to put a comment somewhere saying that `TokenExclusionList` supports both exact matches and regex.

alan20854:
> Nice!
> 
> I'm not exactly sure where this would go, but I think it might be worthwhile to put a comment somewhere saying that `TokenExclusionList` supports both exact matches and regex.

I'll make a note - It probably makes sense to put it in the readme.

internal/clients/nightfall/nightfall.go
```diff
@@ -195,11 +194,16 @@ func createCommentsFromScanResp(
...
 
-func isFindingInTokenExclusionSet(fragment string, tokenExclusionSet *strset.Set) bool {
-	if tokenExclusionSet == nil {
+func isFindingInTokenExclusionList(fragment string, tokenExclusionList []string) bool {
+	if tokenExclusionList == nil {
 		return false
 	}
-	return tokenExclusionSet.Has(fragment)
+	for _, pattern := range tokenExclusionList {
+		if matched, _ := regexp.MatchString(pattern, fragment); matched {
```
bradleymcallister97, line 202: Seems like we should be able to use the [matchRegex](https://github.com/nightfallai/jenkins_test/pull/11/files#diff-d8a0a87ad0f085289ed44c2ba1206b24R361) func from your other diff

alan20854: yup i'll make sure to rebase

henkejosh (approved):

PR #13, opened 2020-08-06
Use Libgit2 to get diff

bradleymcallister97:


.github/workflows/nightfalldlp.yml
```diff
@@ -13,6 +13,18 @@ jobs:
       - name: Download repo (and action since it's currently in a private repo)
         uses: actions/checkout@v2
```
alan20854, line 15: Is this step downloading the repo as well or just installing libgit2?

internal/clients/diffreviewer/github/github_service.go
```diff
@@ -62,38 +63,22 @@ type headCommit struct {
 
 // event represents github event webhook file
 type event struct {
+	Before      string      `json:"before"`
+	BaseRef     string      `json:"base_ref"`
```
alan20854: is this the field that doesn't get populated?

internal/clients/diffreviewer/github/github_service.go
```diff
@@ -176,6 +169,15 @@ func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.C
 	if s.CheckRequest.SHA == "" {
 		s.CheckRequest.SHA = event.HeadCommit.ID
 	}
+	baseRev, ok := os.LookupEnv(BaseRefEnvVar)
+	if !ok || baseRev == "" {
+		s.Logger.Debug(fmt.Sprintf("Could not get environment variable %s so using Before from the event object", BaseRefEnvVar))
+		baseRev = event.Before
```
alan20854, line 174: let's log an error if `event.Before` is empty unless we're guaranteed to have it. On a side note , do you know what would happen for initial commits to new repos?

internal/clients/gitdiff/gitdiff.go
```diff
@@ -0,0 +1,88 @@
...
+	repo, err := c.Libgit.Clone(repoURL, c.RepoFilePath)
+	if err != nil {
+		return nil, err
+	}
+
+	diff, err := c.Libgit.DiffRevToRev(repo, baseRev, headRev)
+	if err != nil {
+		return nil, err
+	}
+
```
alan20854, line 40: what's the difference between a diff and a patch?

bradleymcallister97: The patch is just the format of string that we want to get the diff in. Patch gives us the reguar diff that we get with git diff. The other option I was looking at was Raw but that just gives like a summary of the files changed.

Patch
```
diff --git a/clients/cassandra.go b/clients/cassandra.go
index 94cbad2..ae43ee5 100644
--- a/clients/cassandra.go
+++ b/clients/cassandra.go
@@ -44,8 +44,8 @@ func NewCassandra(
 		cluster.DisableInitialHostLookup = true
 	}

```
Raw
```
:100644 100644 94cbad2... ae43ee5... M	clients/cassandra.go
:100644 100644 05b045a... 3ae0772... M	clients/kafka.go
:100644 100644 9a55398... 182ad8c... M	config/config.go
:100644 100644 5db55b5... 739afc5... M	config/config_test.go
```

internal/clients/diffreviewer/github/github_service.go
bradleymcallister97: I don't think we want an error because if `BaseRefEnvVar` is empty we can still get the `Before` commit hash. This happens for when the action is run on a push.

internal/clients/diffreviewer/github/github_service.go
bradleymcallister97: Yep, will remove!

.github/workflows/nightfalldlp.yml
bradleymcallister97: This just downloads the C version of libgit2, this is required for the `libgit2/git2go` (go version of libgit) to work

alan20854: got it. I suppose the plan is to put this step in our action so users don't have to once we have this tested?

alan20854 (approved):

internal/clients/diffreviewer/github/github_service.go
```diff
@@ -116,18 +100,26 @@ type CheckRequest struct {
 	Name string `json:"name,omitempty"`
 }
 
+type RepoParams struct {
```
henkejosh, line 103: nit - let's add ac omment

internal/clients/diffreviewer/github/github_service.go
```diff
@@ -116,18 +100,26 @@ type CheckRequest struct {
...
+	GitURL  string
+	BaseRev string
```
henkejosh, line 105: nit - should this be `BaseRef`?

internal/clients/gitdiff/gitdiff.go
```diff
@@ -0,0 +1,88 @@
...
+func NewClient(accessToken, repoFilePath string) *Client {
+	return &Client{
+		Libgit:       NewLibgit(accessToken),
+		RepoFilePath: repoFilePath,
+	}
+}
+
+// GetDiff gets the diff from the base to the head on the given repo
+func (c *Client) GetDiff(baseRev, headRev, repoURL string, diffOpts *DiffOptions) ([]*diffreviewer.FileDiff, error) {
+	repo, err := c.Libgit.Clone(repoURL, c.RepoFilePath)
```
henkejosh, line 31: what happens here for giant repos?

internal/clients/diffreviewer/github/github_service.go
bradleymcallister97: The reason I called it `BaseRev` is because depending on which action triggered the action pr or push the value will be retrieved from somewhere different. If the action is a pr then the value is `BaseRef`, if the action is push then the value is `Before` which is the hash of the last time the workflow was run on the branch. So used the name `BaseRev` to be more general.

internal/clients/gitdiff/gitdiff.go
bradleymcallister97: Since the clone also downloads all the history of the repo for large repos it takes a while. I am investigating using a shallow repo and fetching the commits we need to compute the diff.

PR #14, opened 2020-08-07
NIGH-72 Handle API Platform Rate Limiting

alan20854:
https://nightfall.atlassian.net/jira/software/projects/NIGH/boards/19?selectedIssue=NIGH-72

Doing an exponential backoff (x2) from a delay of 1 second when we get 429 Too Many Requests back from the API `/scan` endpoint. 

This does not cover 500 errors passed down from the Amazon API Gateway 429 exceptions, so we currently will not resend the scan request on a 500.


internal/clients/nightfall/nightfall_test.go
```diff
@@ -134,13 +129,72 @@ func (n *nightfallTestSuite) TestScan() {
...
+		}
 
-	mockAPIClient.EXPECT().ScanAPI().Return(mockScanAPI)
-	mockScanAPI.EXPECT().ScanPayload(gomock.Any(), gomock.AssignableToTypeOf(expectedScanReq)).Return(expectedScanResp, nil, nil)
+		numRetries := tt.haveNumReqAttempts
+		if tt.haveSuccess {
+			numRetries--
+		}
+		mockAPIClient.EXPECT().ScanAPI().Return(mockScanAPI)
+		// TODO find out how to send back 429 responses for initial tries
```
alan20854: running into the same issue as `GetRawBySha`, where the http client is nested in the `APIClient`

alan20854:
WIP Tests are currently in progress

internal/clients/nightfall/nightfall.go
```diff
@@ -39,6 +48,7 @@ type Client struct {
 	APIKey             string
 	Detectors          []*nightfallAPI.Detector
 	MaxNumberRoutines  int
+	InitialDelay       time.Duration
```
alan20854: added this field here so it could be configurable in our tests (instead of having to wait 30+ seconds on the max retries failure test)

internal/clients/nightfall/nightfall.go
```diff
@@ -262,19 +283,53 @@ func (n *Client) scanAllContent(ctx context.Context, logger logger.Logger, cts [
...
+				delay = delay * 2 // exponential back off
+				continue
+			}
+			logger.Error(
+				fmt.Sprintf(
+					"Error from Nightfall API, unable to successfully scan %d items",
+					len(request.Payload.Items),
+				),
+			)
+			logger.Error(err.Error())
```
alan20854: rm

internal/mocks/clients/http_client.go
```diff
@@ -1,50 +0,0 @@
-// Code generated by MockGen. DO NOT EDIT.
```
alan20854, line 1: unused

internal/clients/nightfall/nightfall_api.go
```diff
@@ -8,17 +8,23 @@ import (
...
 	return &APIClient{
 		APIClient: nightfallAPI.NewAPIClient(APIConfig),
+		ScanApi:   scanAPI,
 	}
 }
 
 // ScanAPI gets the ScanApi from the api client
 func (c *APIClient) ScanAPI() nightfallintf.NightfallScanAPI {
-	return c.APIClient.ScanApi
+	return c.ScanApi
```
bradleymcallister97: Just curious what is the difference between using the mock  `ScanApi` passed into `NewAPIClient` as opposed to using a mock nightfallAPI client and having the mock `ScanAPI()` func return the mock `ScanApi`

alan20854: The return val of the mock nightfallAPI client's `ScanAPI()` func has type `nightfallintf.NightfallScanAPI`, which doesn't have the mock functions like `EXPECT`

internal/clients/nightfall/nightfall_api.go
```diff
@@ -8,17 +8,23 @@ import (
 // APIClient is a wrapper around nightfallAPI.APIClient
 type APIClient struct {
 	APIClient *nightfallAPI.APIClient
+	ScanApi   nightfallintf.NightfallScanAPI
 }
 
 // NewAPIClient generates a nightfallAPI client
-func NewAPIClient() *APIClient {
+func NewAPIClient(scanAPI nightfallintf.NightfallScanAPI) *APIClient {
```
henkejosh: this approach seems a bit strange to me. Are we sure we need this `ScanAPI` field both behind APIClient and adjacent to it - or can we just move the `ScanPayload` method so that `APIClient` is its receiver?

Haven't played much w/ it or done much testing but I mean something like this: https://github.com/nightfallai/nightfall_cli/pull/26/files

internal/clients/nightfall/nightfall.go
```diff
@@ -30,6 +32,13 @@ const (
 
 	defaultTimeout           = time.Minute * 20
 	MaxConcurrentRoutinesCap = 50
+	maxRetries               = 5
```
henkejosh: nit - do you mind adding comments for each of these vars? We should have caught this in previous PR's for defaultTimeout and MaxConcurrentRoutinesCap. Since these are configuration related and pretty important to the functioning of htis client, it's always good to be extra clear about what's going on.

internal/clients/nightfall/nightfall.go
```diff
@@ -262,19 +283,52 @@ func (n *Client) scanAllContent(ctx context.Context, logger logger.Logger, cts [
...
 	newCtx := context.WithValue(ctx, nightfallAPI.ContextAPIKey, APIKey)
 	request := n.createScanRequest(items)
-	resp, _, err := n.APIClient.ScanAPI().ScanPayload(newCtx, request)
-	if err != nil {
-		logger.Error(fmt.Sprintf("Error from Nightfall API, unable to successfully scan %d items", len(items)))
-		return nil, err
+	return n.makeScanRequest(newCtx, logger, request)
+}
+
+func (n *Client) makeScanRequest(
```
henkejosh: nit - can we update this name to fully reflect its functionality, e.g. `makeScanRequestWithRetries` or something?

internal/clients/nightfall/nightfall.go
```diff
@@ -262,19 +283,52 @@ func (n *Client) scanAllContent(ctx context.Context, logger logger.Logger, cts [
...
+	ctx context.Context,
+	logger logger.Logger,
+	request nightfallAPI.ScanRequest,
+) ([][]nightfallAPI.ScanResponse, error) {
+	delay := n.InitialRetryDelay
+	for i := 0; i < maxRetries; i++ {
+		resp, httpResp, err := n.APIClient.ScanPayload(ctx, request)
+		if err != nil {
+			if httpResp.StatusCode == http.StatusTooManyRequests {
```
henkejosh: just confirming - I know you had mentioned receiving 500 status codes too. Is it possible to get any other status codes when we're being rate limited or should this 1 cover it?

internal/clients/nightfall/nightfall_test.go
```diff
@@ -118,26 +118,92 @@ func (n *nightfallTestSuite) TestScan() {
 	}
 
 	expectedScanReq := createScanReq(detectors, items)
-	expectedScanResp := [][]nightfallAPI.ScanResponse{
-		{},
+	mockAPIClient.EXPECT().ScanPayload(gomock.Any(), gomock.AssignableToTypeOf(expectedScanReq)).Return(expectedScanResponse, nil, nil)
```
henkejosh: why are we only checking the type of the 2nd arg passed to `ScanPayload`? Can't we just pass in the actual `expectedScanReq`?

internal/clients/nightfall/nightfall_test.go
```diff
@@ -118,26 +118,92 @@ func (n *nightfallTestSuite) TestScan() {
...
+				DoAndReturn(
+					func(
+						ctx context.Context,
+						scanReq nightfallAPI.ScanRequest,
+					) ([][]nightfallAPI.ScanResponse, *http.Response, error) {
+						return [][]nightfallAPI.ScanResponse{nil}, expectedTooManyRequestsHTTPResponse, nightfall.ErrMaxScanRetries
+					})
+		}
+		if tt.haveSuccess {
+			mockAPIClient.EXPECT().ScanPayload(gomock.Any(), gomock.AssignableToTypeOf(expectedScanReq)).
```
henkejosh: same as above question - any specific reason we're  only checking the types for all of these mock calls?

internal/clients/nightfall/nightfall.go
alan20854: Nope this is the only status code we receive when we get rate limited from the API Platform. 

When the API Platform gets rate limit errors from the Amazon API Gateway, those get masked as 500s, so I'm not too sure if we want to include 500s in this condition, which may also re-attempt making the request on other unrelated 500 errors.

henkejosh: ah I see. So if the API Platform rate limits us, we get that status code. If there are rate limit issues b/w the API Platform (i.e. Lambda) and the Gateway (i.e. Metadata service holding the companyID's), we get a 500. Makes sense!

internal/clients/nightfall/nightfall_test.go
alan20854: Yea it's because in the `Scan` method, there's a `createScanRequest` call that updates the request with the addition of the detectors, so if we pass in the actual `expectedScanReq`, we get this test failure:

```
Got: {[{CREDIT_CARD_NUMBER} {PHONE_NUMBER}] {[this is a string this has a credit card number 4916-6734-7572-5015 tom cruise!!!!!!]}}
Want: is equal to {[{} {}] {[this is a string this has a credit card number 4916-6734-7572-5015 tom cruise!!!!!!]}}
```

Do you have any suggestions on how to resolve this?

internal/clients/nightfall/nightfall_test.go
```diff
@@ -118,26 +117,93 @@ func (n *nightfallTestSuite) TestScan() {
...
+			numRetries--
+		}
+		for i := 0; i < numRetries; i++ {
+			mockAPIClient.EXPECT().ScanPayload(gomock.Any(), gomock.AssignableToTypeOf(expectedScanReq)).
+				DoAndReturn(
+					func(
+						ctx context.Context,
+						scanReq nightfallAPI.ScanRequest,
+					) ([][]nightfallAPI.ScanResponse, *http.Response, error) {
+						return [][]nightfallAPI.ScanResponse{nil}, expectedTooManyRequestsHTTPResponse, nightfall.ErrMaxScanRetries
```
bradleymcallister97: nit - do we need to use `DoAndReturn` or can we just use `Return`, just trying to simplify this

internal/clients/nightfall/nightfall_test.go
bradleymcallister97: I believe to get this to work we just need a function like the func `createScanRequest` in `nightfall.go` which takes in the detectors and items to scan and returns a `nightfallAPI.ScanRequest`

alan20854: Isn't that what we're doing on line 120?

alan20854: It's weird how the mock client does expect there to be two detectors, but the values are empty.

bradleymcallister97: oh I see, hmmm yea strange

alan20854: Is there a specific reason why detectors is of type `[]*nightfallAPI.Detector`? If I remove the pointer, the tests are now able to read the detector values

internal/clients/nightfall/nightfall_test.go
```diff
@@ -118,33 +117,94 @@ func (n *nightfallTestSuite) TestScan() {
...
+				fmt.Sprintf("Did not get error from %s test", tt.desc),
+			)
+		}
+		n.Equal(tt.wantResponse, resp, fmt.Sprintf("Received incorrect response from Scan in %s test", tt.desc))
+	}
 }
 
 func createScanReq(dets []*nightfallAPI.Detector, items []string) nightfallAPI.ScanRequest {
 	detectors := make([]nightfallAPI.ScanRequestDetectors, 0, len(dets))
-	for d := range dets {
```
alan20854, line 145: this was why we weren't able to read the detectors in the tests

henkejosh: ah don't we have an identical function elsewhere?

I think we do have (at least) 2 and if so let's kill one of them

alan20854: Yea `createScanRequest` in `nightfall.go` has the same functionality with a slight difference in implementation. Are you suggesting we make that one public to access in our test?

henkejosh: Yup it seems like these 2 funcs are basically doing the same thing, so we're currently managing this functionality in 2 places - which has led to us each having to solve this bug in the last 24 hours.

If we need that func for our tests, it probably makes sense to just export it

alan20854: gotcha i'll delete the one in the tests

internal/clients/nightfall/nightfall.go
```diff
@@ -262,19 +286,52 @@ func (n *Client) scanAllContent(ctx context.Context, logger logger.Logger, cts [
...
+func (n *Client) makeScanRequestWithRetries(
+	ctx context.Context,
+	logger logger.Logger,
+	request nightfallAPI.ScanRequest,
+) ([][]nightfallAPI.ScanResponse, error) {
+	delay := n.InitialRetryDelay
+	for i := 0; i < maxRetries; i++ {
+		resp, httpResp, err := n.APIClient.ScanPayload(ctx, request)
+		if err != nil {
+			if httpResp != nil && httpResp.StatusCode == http.StatusTooManyRequests {
```
henkejosh, line 312: can this httpResp be nil?

internal/clients/nightfall/nightfall_test.go
```diff
@@ -89,8 +90,7 @@ func (n *nightfallTestSuite) TestReviewDiff() {
 	expectedComments := []*diffreviewer.Comment{&c, &c, &c}
 
 	for i := 0; i < numScanReq; i++ {
-		mockAPIClient.EXPECT().ScanAPI().Return(mockScanAPI)
-		mockScanAPI.EXPECT().ScanPayload(gomock.Any(), gomock.AssignableToTypeOf(nightfallAPI.ScanRequest{})).Return(scanResp, nil, nil)
+		mockAPIClient.EXPECT().ScanPayload(gomock.Any(), gomock.AssignableToTypeOf(nightfallAPI.ScanRequest{})).Return(expectedScanResponse, nil, nil)
```
henkejosh: lets assert on the actual object, not its type (arg 2)

internal/clients/nightfall/nightfall_test.go
```diff
@@ -117,41 +116,88 @@ func (n *nightfallTestSuite) TestScan() {
...
+			APIClient:         mockAPIClient,
+			Detectors:         detectors,
+			InitialRetryDelay: time.Millisecond,
+		}
+
+		numRetries := tt.haveNumReqAttempts
+		if tt.haveSuccess {
+			numRetries--
+		}
+		for i := 0; i < numRetries; i++ {
```
henkejosh: not sure if this is possible - but can we use [.Times()](https://godoc.org/github.com/golang/mock/gomock#Call.Times) instead of this for loop? Seems like it would simplify this code a bit and allow us to put the `wantNumRetries` value into the test structs instead of calculating it on the fly like on lines 178-181.

What do you think?


Separately, since there's so much conditional logic to check the validity of each test in this test table (and not a ton of shared code), is it cleaner to use subtests? Curious for your thoughts here too.

internal/clients/nightfall/nightfall.go
alan20854: yup there's an error case that returns a nil httpResp

bradleymcallister97 (approved):

internal/clients/nightfall/nightfall_test.go
```diff
@@ -88,70 +99,143 @@ func (n *nightfallTestSuite) TestReviewDiff() {
...
+		if len(totalItems) < startIndex+maxItemsForAPIReq {
+			endIndex = len(totalItems)
+		} else {
+			endIndex = startIndex + maxItemsForAPIReq
+		}
+		scanRequestItems[i] = totalItems[startIndex:endIndex]
+	}
+
+	for i := 0; i < numScanReq; i++ {
+		expectedScanReq := client.CreateScanRequest(scanRequestItems[i])
```
henkejosh: these loops have the exact same conditions - can we do all this logic inside just one of the loops?

henkejosh (approved):

PR #28, opened 2020-08-14
NIGH-89 Ignore config.json

alan20854:
Always adding .nightfalldlp/config.json to the `FileExclusionList`


internal/nightfallconfig/nightfall_config.go
```diff
@@ -53,6 +54,6 @@ func GetNightfallConfigFile(workspacePath, fileName string) (*NightfallConfigFil
 	if nightfallConfig.MaxNumberRoutines == 0 {
 		nightfallConfig.MaxNumberRoutines = defaultMaxNumberRoutines
 	}
-
+	nightfallConfig.FileExclusionList = append(nightfallConfig.FileExclusionList, nightfallConfigFilename)
```
bradleymcallister97, line 57: Could it be possible for `nightfallConfig.FileExclusionList` to be `nil`? if so would this append fail?

alan20854: The append will still work. The `fileExclusionList` in our test config is also currently nil, so this should be fine.

https://play.golang.org/p/CvAWyyZQpr3

bradleymcallister97: Cool!

bradleymcallister97 (approved):

PR #29, opened 2020-08-17
[NIGH-87] Move bash to go

bradleymcallister97:
Ticket: https://nightfall.atlassian.net/browse/NIGH-87


internal/clients/gitdiff/gitdiff.go
```diff
@@ -0,0 +1,66 @@
...
+		return "", err
+	}
+
+	var diffCmd *exec.Cmd
+	if gd.BaseBranch != "" {
+		err = exec.Command("git", "fetch", "origin", gd.BaseBranch).Run()
+		if err != nil {
+			return "", err
+		}
+		diffCmd = exec.Command("git", "diff", fmt.Sprintf("origin/%s", gd.BaseBranch))
```
bradleymcallister97, line 36: Command `git diff origin/${base branch}`. The checkout action checks out the branch for which the PR is being made for (branch with code changes).

Not sure why but `git diff origin/${base branch} ${SHA}` would fail with `128` error code

internal/clients/gitdiff/gitdiff.go
```diff
@@ -0,0 +1,66 @@
...
+	} else if gd.BaseSHA == "" || gd.BaseSHA == unknownCommitHash {
+		err = exec.Command("git", "fetch", "origin", gd.Head, "--depth=2").Run()
+		if err != nil {
+			return "", err
+		}
+		diffCmd = exec.Command("git", "diff", "HEAD^", "HEAD")
```
bradleymcallister97: This is comparing the current commit with the previous commit. The same output as the previously used `git show format='' ${SHA}`

internal/clients/gitdiff/gitdiff.go
alan20854: Comparing this to the bash script, does using fetch depth=1 also cause this error?

internal/clients/gitdiff/gitdiff.go
```diff
@@ -0,0 +1,66 @@
...
+
+	var diffCmd *exec.Cmd
+	if gd.BaseBranch != "" {
+		err = exec.Command("git", "fetch", "origin", gd.BaseBranch).Run()
+		if err != nil {
+			return "", err
+		}
+		diffCmd = exec.Command("git", "diff", fmt.Sprintf("origin/%s", gd.BaseBranch))
+	} else if gd.BaseSHA == "" || gd.BaseSHA == unknownCommitHash {
+		err = exec.Command("git", "fetch", "origin", gd.Head, "--depth=2").Run()
```
alan20854, line 39: can you explain what's different between `depth=1` and `depth=2`?

internal/clients/gitdiff/gitdiff.go
alan20854: Just checking to make sure this still works for initial commits on new repos?

internal/clients/gitdiff/gitdiff.go
bradleymcallister97: The reason I have depth=2 here is that I want the 2 most recent commits for the branch so I can compute the diff with `git diff HEAD^ HEAD`.

alan20854: ah got it

internal/clients/gitdiff/gitdiff.go
bradleymcallister97: Good point, when fetching with `--depth=1` it still works, this probably better for performance instead of fetching the entire branch.

internal/clients/gitdiff/gitdiff.go
bradleymcallister97: Very good point will need to check that out. I assume this will fail since it is trying to use 2 commits and the branch will only have 1.

bradleymcallister97: I have updated this case to use `git show ${SHA} --format=`. `--format=` is the same as `--format=""`, for some reason the command was failing when the `""` was used as the format.

internal/clients/gitdiff/gitdiff.go
```diff
@@ -0,0 +1,66 @@
...
+
+	var diffCmd *exec.Cmd
+	if gd.BaseBranch != "" {
+		err = exec.Command("git", "fetch", "origin", gd.BaseBranch, "--depth=1").Run()
+		if err != nil {
+			return "", err
+		}
+		diffCmd = exec.Command("git", "diff", fmt.Sprintf("origin/%s", gd.BaseBranch))
+	} else if gd.BaseSHA == "" || gd.BaseSHA == unknownCommitHash {
+		err = exec.Command("git", "fetch", "origin", gd.Head, "--depth=2").Run()
```
bradleymcallister97, line 39: Setting depth=2 because if the actions is running on the push of a new branch (just added to remote) we only want to get the diff of its first commit. If depth=1 then the entire repo is stored at the first commit and thus we would end up getting the entire repo as the diff.

alan20854: Aside from the initial push to an new empty repository, would there be other instances that would enter this condition block? 

I'm curious as to whether the BaseSHA can be ""/`unknownCommitHash` on the push of a new branch for a repo with existing commits on the `master` branch, since this condition was initially added just for the purpose of the first push to an empty repo.

bradleymcallister97: On the first push for any new branch will have the `""/unknownCommitHash`, since the action has no before commit for it.

alan20854 (approved):

internal/clients/gitdiff/gitdiff.go
```diff
@@ -0,0 +1,66 @@
...
+
+// GetDiff uses the command line to compute the diff
+func (gd *GitDiff) GetDiff() (string, error) {
+	err := os.Chdir(gd.WorkDir)
+	if err != nil {
+		return "", err
+	}
+
+	var diffCmd *exec.Cmd
+	if gd.BaseBranch != "" {
```
henkejosh: let's use a switch for this complex branching logic

internal/clients/gitdiff/gitdiff.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+		diffCmd = exec.Command("git", "show", gd.Head, "--format=")
+	} else {
+		err = exec.Command("git", "fetch", "origin", gd.BaseSHA, "--depth=1").Run()
+		if err != nil {
+			return "", err
+		}
+		diffCmd = exec.Command("git", "diff", gd.BaseSHA, gd.Head)
+	}
+	reader, err := diffCmd.StdoutPipe()
```
henkejosh, line 54: do we need to close this reader or anything liek that?

internal/clients/gitdiff/gitdiff.go
```diff
@@ -0,0 +1,66 @@
...
+	err = diffCmd.Start()
+	if err != nil {
+		return "", nil
+	}
+	buf := new(strings.Builder)
+	_, err = io.Copy(buf, reader)
+	if err != nil {
+		return "", err
+	}
+	err = diffCmd.Wait()
```
henkejosh, line 69: what does wait do?

internal/clients/gitdiff/gitdiff.go
```diff
@@ -0,0 +1,66 @@
...
+			return "", err
+		}
+		diffCmd = exec.Command("git", "diff", fmt.Sprintf("origin/%s", gd.BaseBranch))
+	} else if gd.BaseSHA == "" || gd.BaseSHA == unknownCommitHash {
+		err = exec.Command("git", "fetch", "origin", gd.Head, "--depth=2").Run()
+		if err != nil {
+			return "", err
+		}
+		diffCmd = exec.Command("git", "show", gd.Head, "--format=")
+	} else {
```
henkejosh: think we could also add comments explaining what each of these cases are/what the diff fetching logic is?

internal/clients/gitdiff/gitdiff.go
bradleymcallister97: The [`Wait()`](https://golang.org/pkg/os/exec/#Cmd.Wait) waits for the command to finish running. Due to the `Copy()` I do not believe that is waits for long or any time, but it does provide an error if the command fails (returns non zero code).

internal/clients/gitdiff/gitdiff.go
```diff
@@ -0,0 +1,73 @@
...
+		// use current commit SHA and previous action run commit SHA for diff
+		err = exec.Command("git", "fetch", "origin", gd.BaseSHA, "--depth=1").Run()
+		if err != nil {
+			return "", err
+		}
+		diffCmd = exec.Command("git", "diff", gd.BaseSHA, gd.Head)
+	}
+
+	reader, err := diffCmd.StdoutPipe()
+	defer reader.Close()
```
henkejosh: Often if there's an error the `reader` is never opened. Just want to confirm - will this cause an error if `diffCmd.StdoutPipe()` returns an error, or are we OK?

internal/clients/gitdiff/gitdiff.go
henkejosh: gotcha. Why are we `Start()`ing and then `Wait()`ing here, whereas other places we just call `Run()` (which I assume is synchronous)? Is this the only way we can redirect the output into our buffer for some reason?

bradleymcallister97: Yea, this is the only way I can see where we can get a reader for the result. For the others we don't care about the output only about if there is an error so that is why I am using `Run()`

internal/clients/gitdiff/gitdiff.go
bradleymcallister97: Gotcha, I can just do the defer after the error check

henkejosh (approved):

PR #32, opened 2020-08-21
Update repo name

henkejosh:


.github/workflows/nightfalldlp.yml
```diff
@@ -32,7 +32,7 @@ jobs:
 
       - name: Setup nightfalldlp
         run: |
-          go install ./cmd/nightfalldlp
+          go build -a -installsuffix cgo -o nightfall_code_scanner ./cmd/nightfalldlp/
```
henkejosh, line 35: updated this to mimic the dockerfile to ensure it isn't broken

alan20854: can you remind me what this command actually does in comparison to the install?

alan20854 (approved):

bradleymcallister97 (approved):

.github/workflows/nightfalldlp.yml
henkejosh: go build builds the package and puts it in the current working dir. go install builds the package and moves it to `GOPATH/bin` so it can be used as a command (i.e. builds and installs it).

PR #35, opened 2020-08-24
NIGH-110 Circle Service Load Config + Logger

alan20854:
Added LoadConfig and Logger for CircleCi Service

### DIFF
--- a/internal/clients/diffreviewer/circleci/circleci_service.go
+++ b/internal/clients/diffreviewer/circleci/circleci_service.go
@@ -46,17 +46,17 @@ func (s *Service) GetLogger() logger.Logger {
 func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.Config, error) {
 	s.Logger.Debug("Loading configuration")
 	workspacePath, ok := os.LookupEnv(WorkspacePathEnvVar)
-	if !ok && workspacePath != "" {
+	if !ok || workspacePath == "" {
 		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", WorkspacePathEnvVar))
 		return nil, errors.New("missing env var for workspace path")
 	}
 	commitSha, ok := os.LookupEnv(CircleCurrentCommitShaEnvVar)
-	if !ok && commitSha != "" {
+	if !ok || commitSha == "" {
 		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", CircleCurrentCommitShaEnvVar))
 		return nil, errors.New("missing env var for commit sha")
 	}
 	beforeCommitSha, ok := os.LookupEnv(CircleBeforeCommitEnvVar)
-	if !ok && beforeCommitSha != "" {
+	if !ok || beforeCommitSha == "" {
 		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", CircleBeforeCommitEnvVar))
 		return nil, errors.New("missing env var for prev commit sha")
 	}
@@ -72,7 +72,7 @@ func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.C
 		return nil, err
 	}
 	nightfallAPIKey, ok := os.LookupEnv(NightfallAPIKeyEnvVar)
-	if !ok && nightfallAPIKey != "" {
+	if !ok || nightfallAPIKey == "" {
 		s.Logger.Error(fmt.Sprintf("Error getting Nightfall API key. Ensure you have %s set in the Github secrets of the repo", NightfallAPIKeyEnvVar))
 		return nil, errors.New("missing env var for nightfall api key")
 	}
--- a/internal/clients/diffreviewer/circleci/circleci_service.go
+++ b/internal/clients/diffreviewer/circleci/circleci_service.go
@@ -15,13 +15,13 @@ import (
 )
 
 const (
-	WorkspacePathEnvVar        = "GITHUB_WORKSPACE"
-	NightfallAPIKeyEnvVar      = "NIGHTFALL_API_KEY"
-	CircleRepoNameEnvVar       = "CIRCLE_PROJECT_REPONAME"
-	CircleOwnerNameEnvVar      = "CIRCLE_PROJECT_USERNAME"
-	CircleCommitShaEnvVar      = "CIRCLE_SHA1"
-	CircleBeforeCommitEnvVar   = "EVENT_BEFORE"
-	CirclePullRequestUrlEnvVar = "CIRCLE_PULL_REQUEST"
+	WorkspacePathEnvVar          = "GITHUB_WORKSPACE"
+	NightfallAPIKeyEnvVar        = "NIGHTFALL_API_KEY"
+	CircleRepoNameEnvVar         = "CIRCLE_PROJECT_REPONAME"
+	CircleOwnerNameEnvVar        = "CIRCLE_PROJECT_USERNAME"
+	CircleCurrentCommitShaEnvVar = "CIRCLE_SHA1"
+	CircleBeforeCommitEnvVar     = "EVENT_BEFORE"
+	CirclePullRequestUrlEnvVar   = "CIRCLE_PULL_REQUEST"
 )
 
 // Service contains the github client that makes Github api calls
@@ -50,9 +50,9 @@ func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.C
 		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", WorkspacePathEnvVar))
 		return nil, errors.New("Missing env var for workspace path")
 	}
-	commitSha, ok := os.LookupEnv(CircleCommitShaEnvVar)
+	commitSha, ok := os.LookupEnv(CircleCurrentCommitShaEnvVar)
 	if !ok {
-		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", CircleCommitShaEnvVar))
+		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", CircleCurrentCommitShaEnvVar))
 		return nil, errors.New("Missing env var for commit sha")
 	}
 	beforeCommitSha, ok := os.LookupEnv(CircleBeforeCommitEnvVar)
--- a/internal/clients/diffreviewer/circleci/circleci_service.go
+++ b/internal/clients/diffreviewer/circleci/circleci_service.go
@@ -56,10 +56,6 @@ func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.C
 		return nil, errors.New("missing env var for commit sha")
 	}
 	beforeCommitSha, ok := os.LookupEnv(CircleBeforeCommitEnvVar)
-	if !ok || beforeCommitSha == "" {
-		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", CircleBeforeCommitEnvVar))
-		return nil, errors.New("missing env var for prev commit sha")
-	}
 	s.GitDiff = &gitdiff.GitDiff{
 		WorkDir:    workspacePath,
 		BaseBranch: "master",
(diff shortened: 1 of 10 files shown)


internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -0,0 +1,106 @@
...
+	circlelogger "github.com/nightfallai/nightfall_code_scanner/internal/clients/logger/circle_logger"
+	"github.com/nightfallai/nightfall_code_scanner/internal/clients/nightfall"
+	"github.com/nightfallai/nightfall_code_scanner/internal/interfaces/gitdiffintf"
+	"github.com/nightfallai/nightfall_code_scanner/internal/nightfallconfig"
+)
+
+const (
+	WorkspacePathEnvVar        = "GITHUB_WORKSPACE"
+	NightfallAPIKeyEnvVar      = "NIGHTFALL_API_KEY"
+	CircleRepoNameEnvVar       = "CIRCLE_PROJECT_REPONAME"
```
alan20854: may or may not need some of these env vars based on what we need for the github Issues API

internal/clients/logger/circle_logger/circle_logger.go
```diff
@@ -0,0 +1,47 @@
+package circlelogger
```
bradleymcallister97, line 1: Can we add tests for these? Should be very similar to the `github_logger_test.go`

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -0,0 +1,106 @@
...
+	CircleOwnerNameEnvVar      = "CIRCLE_PROJECT_USERNAME"
+	CircleCommitShaEnvVar      = "CIRCLE_SHA1"
```
bradleymcallister97: I assume this is the current commit that triggered the action? If so can we rename `CircleCommitShaEnvVar` to be more specific like `CircleCurrentCommitShaEnvVar`

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	nightfallConfig, err := nightfallconfig.GetNightfallConfigFile(workspacePath, nightfallConfigFileName)
+	if err != nil {
+		s.Logger.Error("Error getting Nightfall config file. Ensure you have a Nightfall config file located in the root of your repository at .nightfalldlp/config.json with at least one Detector enabled")
+		return nil, err
+	}
+	nightfallAPIKey, ok := os.LookupEnv(NightfallAPIKeyEnvVar)
+	if !ok {
+		s.Logger.Error(fmt.Sprintf("Error getting Nightfall API key. Ensure you have %s set in the Github secrets of the repo", NightfallAPIKeyEnvVar))
+		return nil, errors.New("Missing env var for nightfall api key")
```
bradleymcallister97: nit - I believe we want the errors to start with lower cases (Not sure if we did this for Github Action), but probably good to do it at least for Circle CI. There are also a couple of error messages that start with an upper case above.

bradleymcallister97 (approved):

internal/clients/diffreviewer/circleci/circleci_service.go
alan20854: yup gotcha i'll make sure they all start with lower case

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, errors.New("missing env var for commit sha")
+	}
+	beforeCommitSha, ok := os.LookupEnv(CircleBeforeCommitEnvVar)
+	if !ok {
+		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", CircleBeforeCommitEnvVar))
+		return nil, errors.New("missing env var for prev commit sha")
+	}
+	s.GitDiff = &gitdiff.GitDiff{
+		WorkDir:    workspacePath,
+		BaseBranch: "master", //TODO: look into how to get this instead of hardcoding
```
henkejosh: let's not leave TODO's but instead just create a ticket on the board to figure this out

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -0,0 +1,106 @@
...
+
+// LoadConfig gets all config values from files or environment and creates a config
+func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.Config, error) {
+	s.Logger.Debug("Loading configuration")
+	workspacePath, ok := os.LookupEnv(WorkspacePathEnvVar)
+	if !ok {
+		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", WorkspacePathEnvVar))
+		return nil, errors.New("missing env var for workspace path")
+	}
+	commitSha, ok := os.LookupEnv(CircleCurrentCommitShaEnvVar)
```
henkejosh, line 53: just checking to make sure - the `LookupEnv` `ok` return value is always true if the variable exists, regardless of whether it's empty. So if any of these Env vars cannot have valid empty values and are also not automatically set by circle, it's probably best to do an additional check to make sure the var isn't empty so we can make sure to log a helpful error message instead of just continuing on (e.g.  of `!ok && var != ""`)

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -0,0 +1,106 @@
...
+		BaseSHA:    beforeCommitSha,
+		Head:       commitSha,
+	}
+	nightfallConfig, err := nightfallconfig.GetNightfallConfigFile(workspacePath, nightfallConfigFileName)
+	if err != nil {
+		s.Logger.Error("Error getting Nightfall config file. Ensure you have a Nightfall config file located in the root of your repository at .nightfalldlp/config.json with at least one Detector enabled")
+		return nil, err
+	}
+	nightfallAPIKey, ok := os.LookupEnv(NightfallAPIKeyEnvVar)
+	if !ok {
```
henkejosh: probably need to add an extra `!ok && nightfallAPIKey != ""`, otherwise if they pass in a NIGHTFALL_API_KEY env var to their configuration but forget to actually set that value in their Circle secrets, this will pass

internal/clients/diffreviewer/circleci/circleci_service_test.go
```diff
@@ -0,0 +1,97 @@
...
+		Logger: circleLogger,
+	}
+	return tp
+}
+
+const commitSha = "7b46da6e4d3259b1a1c470ee468e2cb3d9733802"
+const prevCommitSha = "15bf9548d16caff9f398b5aae78a611fc60d55bd"
+const testConfigFileName = "nightfall_test_config.json"
+const excludedCreditCardRegex = "4242-4242-4242-[0-9]{4}"
+const excludedApiToken = "xG0Ct4Wsu3OTcJnE1dFLAQfRgL6b8tIv"
```
henkejosh, line 42: can we add this to the exclusion list in our .nightfalldlp config?

internal/clients/diffreviewer/diffutils/diff_filter.go
```diff
@@ -0,0 +1,46 @@
+package diffutils
+
+import (
+	"strings"
+
+	"github.com/nightfallai/nightfall_code_scanner/internal/clients/diffreviewer"
+)
+
+func FilterFileDiffs(fileDiffs []*diffreviewer.FileDiff) []*diffreviewer.FileDiff {
```
henkejosh, line 9: all of these funcs were copy/pasta from github package right? Nothing net new?

alan20854: yup I just pulled them out into their own package

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -0,0 +1,106 @@
...
+// GetLogger gets the github service logger
+func (s *Service) GetLogger() logger.Logger {
+	return s.Logger
+}
+
+// LoadConfig gets all config values from files or environment and creates a config
+func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.Config, error) {
+	s.Logger.Debug("Loading configuration")
+	workspacePath, ok := os.LookupEnv(WorkspacePathEnvVar)
+	if !ok && workspacePath != "" {
```
henkejosh: sorry I was writing quickly - I think we actually want to test that the workspacePath IS empty here (==). Is there any situation in which this can be empty?

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -0,0 +1,106 @@
...
+		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", WorkspacePathEnvVar))
+		return nil, errors.New("missing env var for workspace path")
+	}
+	commitSha, ok := os.LookupEnv(CircleCurrentCommitShaEnvVar)
+	if !ok && commitSha != "" {
```
henkejosh: same ^

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -0,0 +1,106 @@
...
+		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", CircleCurrentCommitShaEnvVar))
+		return nil, errors.New("missing env var for commit sha")
+	}
+	beforeCommitSha, ok := os.LookupEnv(CircleBeforeCommitEnvVar)
+	if !ok && beforeCommitSha != "" {
```
henkejosh: same ^

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -0,0 +1,106 @@
...
+		BaseSHA:    beforeCommitSha,
+		Head:       commitSha,
+	}
+	nightfallConfig, err := nightfallconfig.GetNightfallConfigFile(workspacePath, nightfallConfigFileName)
+	if err != nil {
+		s.Logger.Error("Error getting Nightfall config file. Ensure you have a Nightfall config file located in the root of your repository at .nightfalldlp/config.json with at least one Detector enabled")
+		return nil, err
+	}
+	nightfallAPIKey, ok := os.LookupEnv(NightfallAPIKeyEnvVar)
+	if !ok && nightfallAPIKey != "" {
```
henkejosh: `==`

internal/clients/diffreviewer/circleci/circleci_service_test.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (c *circleCiTestSuite) AfterTest(suiteName, testName string) {
+	for _, e := range envVars {
+		err := os.Unsetenv(e)
+		c.NoErrorf(err, "Error unsetting var %s", e)
+	}
+}
+
+func (c *circleCiTestSuite) TestLoadConfig() {
```
henkejosh, line 59: since we have the weirdness w/ the Env variables and checking their existence in the LoadConfig method, lets add a test or two to make sure that we return the correct error if they're empty. We don't need to check all of them but at least the NIGHTFALL_API_KEY

internal/clients/diffreviewer/circleci/circleci_service.go
alan20854: shoot my bad haha - and no I don't think so since I'm manually setting it into the environment in the circle orb. I think we should still check that it's not empty however in case it's not there

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -0,0 +1,102 @@
...
+	if !ok || workspacePath == "" {
+		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", WorkspacePathEnvVar))
+		return nil, errors.New("missing env var for workspace path")
+	}
+	commitSha, ok := os.LookupEnv(CircleCurrentCommitShaEnvVar)
+	if !ok || commitSha == "" {
+		s.Logger.Error(fmt.Sprintf("Environment variable %s cannot be found", CircleCurrentCommitShaEnvVar))
+		return nil, errors.New("missing env var for commit sha")
+	}
+	beforeCommitSha, ok := os.LookupEnv(CircleBeforeCommitEnvVar)
```
henkejosh: we don't use this `ok` value, right?

henkejosh (approved):

PR #38, opened 2020-08-28
Add info function to logger interface

alan20854:
Per Louis's suggestion, update the circle workflow output from
Screen Shot 2020-08-28 at 3 21 39 PM
to
Screen Shot 2020-08-28 at 3 21 18 PM

### DIFF
--- a/internal/clients/logger/github_logger/github_logger.go
+++ b/internal/clients/logger/github_logger/github_logger.go
@@ -39,6 +39,11 @@ func (l *GithubLogger) Debug(msg string) {
 	l.log.Printf("%s%s\n", debugPrefix, msg)
 }
 
+// Info logs an info message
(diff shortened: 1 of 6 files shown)


internal/clients/logger/github_logger/github_logger.go
```diff
@@ -39,6 +39,11 @@ func (l *GithubLogger) Debug(msg string) {
 	l.log.Printf("%s%s\n", debugPrefix, msg)
 }
 
+// Info logs an info message
+func (l *GithubLogger) Info(msg string) {
+	l.log.Println(msg)
```
bradleymcallister97, line 44: I am wondering if we want any prefix here. I cannot think of any good prefix, looking at other actions they also just plainly log text so probably fine.

bradleymcallister97 (approved):

PR #39, opened 2020-08-31
Write Non-Duplicated Comments

alan20854:
Adds the functionality to check if a comment has already been posted (same body, path, and line number), and only create the comment if not


internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -199,8 +199,19 @@ func (s *Service) WriteComments(comments []*diffreviewer.Comment) error {
 		return nil
 	}
 	if s.PrDetails.PrNumber != nil {
+		existingComments, _, err := s.GithubClient.PullRequestsService().ListComments(
```
bradleymcallister97, line 202: Is there a way to filter the comments we get? I assume this would also fetch the comments from engineers reviewing the diff

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -259,3 +270,33 @@ func (s *Service) createGithubRepositoryComments(comments []*diffreviewer.Commen
...
+	existingCommentsMap := make(map[prComment]bool, len(existingComments))
+	for _, ec := range existingComments {
+		comment := prComment{
+			Body: *ec.Body,
+			Path: *ec.Path,
+			Line: *ec.Line,
+		}
+		existingCommentsMap[comment] = true
+	}
+	filteredComments := []*github.PullRequestComment{}
```
henkejosh: i imagine the capacity for filteredComments will typically be around the length of `comments` (since hopefully users will fix their sensitive info more often than not), so I think we can specify a slice capacity here

internal/clients/diffreviewer/circleci/circleci_service.go
alan20854: Yea good question. It doesn't look like we can pass in any relevant filters in the `PullRequestListCommentsOptions`, as those filters are just `sort` (created vs updated comment), `direction` (asc vs desc), and `since` (time).

Since the comment is written with a user's personal access token, the comment will appear the same as a user-written one. What I could do is iterate through all the comments we get back, and only retain the one's where the comment body start with `Suspicious content detected`, but I'm not sure if that actually addresses your question (or even cuts down the runtime).

henkejosh (approved):

bradleymcallister97 (approved):

PR #41, opened 2020-09-03
NIGH-153 Fix FilterComments Error

alan20854:
`ec(github.PullRequestComment).Line` does not exist for outdated comments, which was causing an nil dereference error, but `ec.GetLine()` will return 0 if it does not exist.


bradleymcallister97:
Do we need to update test cases to cover this? Maybe some test cases for the func `filterExistingComments`?

henkejosh (approved):

bradleymcallister97 (approved):

PR #42, opened 2020-09-09
NIGH-154 Default Config

alan20854:
https://nightfall.atlassian.net/jira/software/projects/NIGH/boards/19?selectedIssue=NIGH-154

Use default detectors (API_KEY and CRYTOGRAPHIC_TOKEN) when config file is missing/invalid.

Added to the github service as well, in case we want to allow this functionality there too.

### DIFF
--- a/internal/clients/diffreviewer/circleci/circleci_service.go
+++ b/internal/clients/diffreviewer/circleci/circleci_service.go
@@ -8,8 +8,6 @@ import (
 	"strconv"
 	"strings"
 
-	nightfallAPI "github.com/nightfallai/nightfall_go_client/generated"
-
 	"github.com/google/go-github/v31/github"
 	"github.com/nightfallai/nightfall_code_scanner/internal/clients/diffreviewer"
 	"github.com/nightfallai/nightfall_code_scanner/internal/clients/diffreviewer/diffutils"
@@ -21,6 +19,7 @@ import (
 	"github.com/nightfallai/nightfall_code_scanner/internal/interfaces/gitdiffintf"
 	"github.com/nightfallai/nightfall_code_scanner/internal/interfaces/githubintf"
 	"github.com/nightfallai/nightfall_code_scanner/internal/nightfallconfig"
+	nightfallAPI "github.com/nightfallai/nightfall_go_client/generated"
 )
 
 const (
--- a/internal/clients/diffreviewer/circleci/circleci_service.go
+++ b/internal/clients/diffreviewer/circleci/circleci_service.go
@@ -19,7 +19,6 @@ import (
 	"github.com/nightfallai/nightfall_code_scanner/internal/interfaces/gitdiffintf"
 	"github.com/nightfallai/nightfall_code_scanner/internal/interfaces/githubintf"
 	"github.com/nightfallai/nightfall_code_scanner/internal/nightfallconfig"
-	nightfallAPI "github.com/nightfallai/nightfall_go_client/generated"
 )
 
 const (
@@ -41,8 +40,6 @@ const (
 )
 
 var errSensitiveItemsFound = errors.New("potentially sensitive items found")
-var apiKeyDetector = nightfallAPI.API_KEY
-var cryptoKeyDetector = nightfallAPI.CRYPTOGRAPHIC_TOKEN
 
 // Service contains the github client that makes Github api calls
 type Service struct {
--- a/internal/clients/diffreviewer/circleci/circleci_service.go
+++ b/internal/clients/diffreviewer/circleci/circleci_service.go
@@ -8,6 +8,8 @@ import (
 	"strconv"
 	"strings"
 
+	nightfallAPI "github.com/nightfallai/nightfall_go_client/generated"
+
 	"github.com/google/go-github/v31/github"
 	"github.com/nightfallai/nightfall_code_scanner/internal/clients/diffreviewer"
 	"github.com/nightfallai/nightfall_code_scanner/internal/clients/diffreviewer/diffutils"
@@ -40,6 +42,8 @@ const (
 )
 
 var errSensitiveItemsFound = errors.New("potentially sensitive items found")
+var apiKeyDetector = nightfallAPI.API_KEY
+var cryptoKeyDetector = nightfallAPI.CRYPTOGRAPHIC_TOKEN
 
 // Service contains the github client that makes Github api calls
 type Service struct {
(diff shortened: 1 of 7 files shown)


internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -112,7 +118,7 @@ func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.C
 	}
 
 	var maxNumberRoutines int
-	if nightfallConfig.MaxNumberRoutines < nightfall.MaxConcurrentRoutinesCap {
+	if nightfallConfig.MaxNumberRoutines < nightfall.MaxConcurrentRoutinesCap && nightfallConfig.MaxNumberRoutines > 0 {
```
alan20854: do we want to set the default to something lower than the max when there isn't one? How does 20 sound (`nightfall.MaxConcurrentRoutinesCap=50`)

internal/clients/diffreviewer/github/github_service.go
```diff
@@ -191,16 +196,18 @@ func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.C
...
+		}
 	}
 	nightfallAPIKey, ok := os.LookupEnv(NightfallAPIKeyEnvVar)
 	if !ok || nightfallAPIKey == "" {
 		s.Logger.Error(fmt.Sprintf("Error getting Nightfall API key. Ensure you have %s set in the Github secrets of the repo", NightfallAPIKeyEnvVar))
 		return nil, errors.New("Missing env var for nightfall api key")
 	}
 	var maxNumberRoutines int
-	if nightfallConfig.MaxNumberRoutines < nightfall.MaxConcurrentRoutinesCap {
+	if nightfallConfig.MaxNumberRoutines < nightfall.MaxConcurrentRoutinesCap && nightfallConfig.MaxNumberRoutines > 0 {
```
alan20854: same here

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -102,8 +105,10 @@ func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.C
 	}
 	nightfallConfig, err := nightfallconfig.GetNightfallConfigFile(workspacePath, nightfallConfigFileName)
 	if err != nil {
-		s.Logger.Error("Error getting Nightfall config file. Ensure you have a Nightfall config file located in the root of your repository at .nightfalldlp/config.json with at least one Detector enabled")
-		return nil, err
+		s.Logger.Warning("Issue retrieving valid Nightfall config file, using default detectors instead. Ensure you have a Nightfall config file located in the root of your repository at .nightfalldlp/config.json with at least one Detector enabled")
```
bradleymcallister97: Should we list the default detectors in a log?

internal/clients/diffreviewer/github/github_service.go
```diff
@@ -191,16 +194,18 @@ func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.C
 	}
 	nightfallConfig, err := nightfallconfig.GetNightfallConfigFile(workspacePath, nightfallConfigFileName)
 	if err != nil {
-		s.Logger.Error("Error getting Nightfall config file. Ensure you have a Nightfall config file located in the root of your repository at .nightfalldlp/config.json with at least one Detector enabled")
-		return nil, err
+		s.Logger.Warning("Issue retrieving valid Nightfall config file, using default detectors instead. Ensure you have a Nightfall config file located in the root of your repository at .nightfalldlp/config.json with at least one Detector enabled")
+		nightfallConfig = &nightfallconfig.NightfallConfigFileStructure{
```
bradleymcallister97: If we want to have a default config for all the CIs we support, then I am wondering if this logic should go inside of `nightfallconfig.GetNightfallConfigFile`, thoughts?

alan20854: Good idea, that would get rid of the duplicate code

internal/clients/diffreviewer/github/github_service.go
```diff
@@ -203,7 +198,7 @@ func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.C
 	if nightfallConfig.MaxNumberRoutines < nightfall.MaxConcurrentRoutinesCap {
 		maxNumberRoutines = nightfallConfig.MaxNumberRoutines
 	} else {
-		maxNumberRoutines = nightfall.MaxConcurrentRoutinesCap
+		maxNumberRoutines = nightfallconfig.DefaultMaxNumberRoutines
```
bradleymcallister97: Should we also just handle this logic inside of `nightfallconfig.GetNightfallConfigFile`, since this is also the same for all clients?

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -100,27 +99,16 @@ func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.C
...
-		s.Logger.Error("Error getting Nightfall config file. Ensure you have a Nightfall config file located in the root of your repository at .nightfalldlp/config.json with at least one Detector enabled")
-		return nil, err
-	}
+	nightfallConfig := nightfallconfig.GetNightfallConfigFile(workspacePath, nightfallConfigFileName, s.Logger)
 	nightfallAPIKey, ok := os.LookupEnv(NightfallAPIKeyEnvVar)
 	if !ok || nightfallAPIKey == "" {
 		s.Logger.Error(fmt.Sprintf("Error getting Nightfall API key. Ensure you have %s set in the Github secrets of the repo", NightfallAPIKeyEnvVar))
 		return nil, errors.New("missing env var for nightfall api key")
 	}
-
```
alan20854, line 113: moved this logic to nightfall_config

internal/nightfallconfig/nightfall_config.go
```diff
@@ -32,28 +39,40 @@ type Config struct {
...
-		return nil, err
+		logger.Warning(fmt.Sprintf("Error reading nightfall config: %s", err.Error()))
+		logger.Info(defaultDetectorsInfoMessage)
+		return defaultNightfallConfig
 	}
 	var nightfallConfig NightfallConfigFileStructure
 	err = json.Unmarshal(byteValue, &nightfallConfig)
 	if err != nil {
-		return nil, err
+		logger.Warning("Invalid nightfall config")
```
henkejosh: i'm not sure we want to use the default conditions here - if someone misconfigures their config (say they want to scan for SSN and CC numbers), and we just scan all of their code w/ a different condition set I think it could be confusing (most people won't be reading their Action/Circle logs for this stuff).

What do you think? Should we throw an error and force them to fix it or delete it, or should we just move ahead w/ scanning using our default detectors?

internal/nightfallconfig/nightfall_config.go
```diff
@@ -32,28 +39,40 @@ type Config struct {
...
 	}
 	if len(nightfallConfig.Detectors) < 1 {
-		return nil, errors.New("Nightfall config file is missing detectors")
+		logger.Warning("Nightfall config requires at least one detector")
+		logger.Info(defaultDetectorsInfoMessage)
+		return defaultNightfallConfig
 	}
-	if nightfallConfig.MaxNumberRoutines == 0 {
-		nightfallConfig.MaxNumberRoutines = defaultMaxNumberRoutines
+	if nightfallConfig.MaxNumberRoutines > MaxConcurrentRoutinesCap || nightfallConfig.MaxNumberRoutines <= 0 {
```
henkejosh: hmm I think if someone specifies like 100 for the MaxNumberRoutines, we would want to give them as many as possible. If someone specifies 0 or less, we should give them the default. What do you think?

alan20854: Yeah I agree, I'll add an extra condition for that

internal/nightfallconfig/nightfall_config.go
alan20854: Ok that makes sense - since they're already explicitly including a nightfall config, we should expect it to be valid (and error otherwise)

henkejosh (approved):

bradleymcallister97 (approved):

PR #45, opened 2020-10-14
Use Conditions & inspectv2

henkejosh:
- update NightfallConfig to take in conditions instead of detectors
- update .nightfalldlp/config.json file to use Conditions
- update errors, logs, etc. to reference conditions instead of detectors
- use v2 generated types instead of v1


internal/nightfallconfig/nightfall_config.go
```diff
@@ -42,30 +63,26 @@ type Config struct {
...
 		return defaultNightfallConfig, nil
 	}
 	var nightfallConfig NightfallConfigFileStructure
 	err = json.Unmarshal(byteValue, &nightfallConfig)
 	if err != nil {
 		return nil, err
 	}
-	if len(nightfallConfig.Detectors) < 1 {
-		return nil, errors.New("Nightfall config file is missing detectors")
+	if len(nightfallConfig.Conditions) < 1 {
```
PetraZ, line 84: [mark] im not sure, are we planning to accept only conditionSetUUID for these pr scanner

.nightfalldlp/config.json
```diff
@@ -1,9 +1,41 @@
...
-        "CRYPTOGRAPHIC_KEY"
+    "conditions": [
+        {
+            "detector": {
+                "detectorType": "NIGHTFALL_DETECTOR",
+                "nightfallDetector": "CREDIT_CARD_NUMBER",
+                "displayName": "cc num"
+            },
+            "minNumFindings": 1,
+            "minConfidence": "VERY_UNLIKELY"
```
alan20854, line 10: just curious but is there a reason why we want the confidence to be `VERY_UNLIKELY` for each detector?

internal/nightfallconfig/nightfall_config.go
alan20854: +1 I think we should also add a `ConditionSetUUID` field wherever you're changing `Detectors` -> `Conditions`, and update this if condition to:
`if len(nightfallConfig.Conditions) < 1 && nightfallConfig.ConditionSetUUID == nil {`

internal/clients/nightfall/nightfall.go
```diff
@@ -47,7 +47,7 @@ var (
 type Client struct {
 	APIClient          nightfallintf.NightfallAPI
 	APIKey             string
-	Detectors          []*nightfallAPI.Detector
+	Conditions         []*nightfallAPI.Condition
```
alan20854, line 50: either create a struct containing both `[]*nightfallAPI.Condition` and `UUID`, or simply add a `ConditionSetUUID` field in the case where users just input a pre-built condition set uuid

.nightfalldlp/config.json
```diff
@@ -1,9 +1,41 @@
...
+        },
+        {
+            "detector": {
+                "detectorType": "NIGHTFALL_DETECTOR",
+                "nightfallDetector": "CRYPTOGRAPHIC_KEY",
+                "displayName": "crypto key"
+            },
+            "minNumFindings": 1,
+            "minConfidence": "VERY_UNLIKELY"
+        }
```
PetraZ, line 38: similar product question:
do we want to support context rule/ token exclusion rule /inline regex stuff

alan20854: it should already be supported with the v2 go client, seems like they're just not included in our own config for testing

alan20854: Related to this: do we still want to support overall token exclusion now that we have exclusion rules per condition? If so, we'd be filtering out tokens in `tokenExclusionList` below at the very end with all findings aggregated (and filtered) from each condition.

File inclusion/exclusion seems ok to leave as is for now since we don't have that functionality elsewhere.

internal/nightfallconfig/nightfall_config.go
henkejosh: yep good call - i just figured that could be added in a separate PR

.nightfalldlp/config.json
henkejosh: 🤷‍♂️ this file is really just for doing local testing when we push up diffs, so I wanted to see findings regardless of how low their likelihood was. We can bump if needed but I can't imagine us having any issues w/ getting too many findings or something.

.nightfalldlp/config.json
henkejosh: Based on our conversation in standup, ^ we're going to continue to support the global token exclusion/inclusion

alan20854 (approved):

PetraZ (approved):

PR #46, opened 2020-10-16
NIGH-191 Add Condition Set UUID

alan20854:
Add ability for users to scan with a pre-built condition set w/ its UUID


internal/nightfallconfig/nightfall_config.go
```diff
@@ -44,6 +44,7 @@ var defaultNightfallConfig = &NightfallConfigFileStructure{
 
 // NightfallConfigFileStructure struct representation of nightfall config file
 type NightfallConfigFileStructure struct {
+	ConditionSetUUID   string                    `json:"conditionSetUUID"`
```
bradleymcallister97, line 48: Is there a reason this is not `uuid.UUID` instead of `string`?

internal/nightfallconfig/nightfall_config.go
```diff
@@ -81,8 +83,8 @@ func GetNightfallConfigFile(workspacePath, fileName string, logger logger.Logger
 	if err != nil {
 		return nil, err
 	}
-	if len(nightfallConfig.Conditions) < 1 {
-		return nil, errors.New("Nightfall config file is missing Conditions")
+	if nightfallConfig.ConditionSetUUID == "" && len(nightfallConfig.Conditions) < 1 {
```
bradleymcallister97, line 93: Is `ConditionSetUUID` and `Conditions` mutually exclusive? I seem to remember that the call to DSG would fail if we send a request with both `ConditionSetUUID` and `Conditions` set.

alan20854: Yup that's correct. This is handled in the lambda (if we have both when we hit the lambda, the lambda will default to using the `ConditionSetUUID` in the request to DSG).

internal/nightfallconfig/nightfall_config.go
alan20854: The OpenAPI generated ConditionSetUUID type is also a string (uuids aren't supported), so I didn't want to convert this into a `uuid.UUID` when parsing the config.json and then back into a string as the `ScanRequestV2` struct expects upon calling `ScanPayload`.

bradleymcallister97: Gotcha! Should we do any validation to make sure the string is a valid UUID?

alan20854: Yea good question - it might be a good idea to fail here rather than failing downstream (there's validation on the proto when making the request to DSG that will error if it's invalid). I'll add a check here too

henkejosh (approved):

bradleymcallister97 (approved):
Do need to rebase? Since base branch is not master

internal/clients/diffreviewer/circleci/circleci_service.go
```diff
@@ -101,7 +101,13 @@ func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.C
 	}
 	nightfallConfig, err := nightfallconfig.GetNightfallConfigFile(workspacePath, nightfallConfigFileName, s.Logger)
 	if err != nil {
+<<<<<<< HEAD
```
alan20854: oops fix

internal/clients/diffreviewer/github/github_service.go
```diff
@@ -189,7 +189,13 @@ func (s *Service) LoadConfig(nightfallConfigFileName string) (*nightfallconfig.C
 	}
 	nightfallConfig, err := nightfallconfig.GetNightfallConfigFile(workspacePath, nightfallConfigFileName, s.Logger)
 	if err != nil {
+<<<<<<< HEAD
```
alan20854: same

PR #52, opened 2021-03-10
Configure Base URL for Github Client

alan20854:
adds the ability to write to a specified url instead of the default https://api.github.com/

Looking at https://github.com/google/go-github/issues/958, looks like the url needs to be suffixed by `/api/v3/`

### DIFF
--- a/.github/workflows/nightfalldlp.yml
+++ b/.github/workflows/nightfalldlp.yml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - name: Add $GOPATH/bin
         run: |
-          echo ::add-path::$(go env GOPATH)/bin
+          echo $(go env GOPATH)/bin >> $GITHUB_PATH
 
       - name: Download repo
         uses: actions/checkout@v2
--- a/cmd/nightfalldlp/main.go
+++ b/cmd/nightfalldlp/main.go
@@ -17,7 +17,7 @@ const (
 	nightfallConfigFileName = ".nightfalldlp/config.json"
 	githubActionsEnvVar     = "GITHUB_ACTIONS"
 	githubTokenEnvVar       = "GITHUB_TOKEN"
-	githubBaseUrlEnvVar     = "BASE_URL"
+	githubApiBaseUrlEnvVar  = "GITHUB_API_BASE_URL"
 	circleCiEnvVar          = "CIRCLECI"
 )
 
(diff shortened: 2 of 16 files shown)


.github/workflows/nightfalldlp.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     steps:
       - name: Add $GOPATH/bin
         run: |
-          echo ::add-path::$(go env GOPATH)/bin
+          echo $(go env GOPATH)/bin >> $GITHUB_PATH
```
alan20854, line 14: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#adding-a-system-path

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
Screen Shot 2021-03-10 at 11 13 06 AM

alan20854:
How can I test this for enterprise?

cmd/nightfalldlp/main.go
```diff
@@ -17,6 +17,7 @@ const (
 	nightfallConfigFileName = ".nightfalldlp/config.json"
 	githubActionsEnvVar     = "GITHUB_ACTIONS"
 	githubTokenEnvVar       = "GITHUB_TOKEN"
+	githubBaseUrlEnvVar     = "BASE_URL"
```
alan20854: TODO: update readme docs on https://github.com/nightfallai/nightfall_dlp_action and https://github.com/nightfallai/nightfall_circle_orb on how to use this

rssathe: Shouldn't this be GITHUB_API_BASE_URL

rssathe: likewise for the variable Name

cmd/nightfalldlp/main.go
```diff
@@ -87,15 +88,17 @@ func CreateDiffReviewerClient() (diffreviewer.DiffReviewer, error) {
 		if !ok {
 			return nil, fmt.Errorf("could not find required %s environment variable", githubTokenEnvVar)
 		}
-		return github.NewAuthenticatedGithubService(githubToken), nil
+		baseUrl, _ := os.LookupEnv(githubBaseUrlEnvVar)
```
rssathe: Can't you jsut create this once above the switch case?

internal/clients/diffreviewer/github/github_client.go
```diff
@@ -14,13 +16,24 @@ type Client struct {
...
 	ctx := context.Background()
 	ts := oauth2.StaticTokenSource(
 		&oauth2.Token{AccessToken: token},
 	)
 	tc := oauth2.NewClient(ctx, ts)
 	githubClient := github.NewClient(tc)
+	// for enterprise
+	if baseUrl != "" {
+		u, _ := url.Parse(baseUrl)
+		if !strings.HasSuffix(u.Path, "/") {
```
rssathe, line 29: is it not worth just overwriting the path entirely with fmt.String("%v%v", u.HostName, githubAPIPath ). Do we want them to customize this. Seems like we should ensure it's always v3?

rssathe (approved):
Nits

internal/clients/diffreviewer/github/github_client.go
alan20854: if they don't specify a protocol, only `u.Path` has a value, the full path including the suffixes, and u.Host is empty. If we want to go the suggested route ^, we'd have to specifically tell users to prepend the protocol when setting their base url.

PR #53, opened 2021-03-11
fix env var name

alan20854:
https://docs.github.com/en/actions/reference/environment-variables#naming-conventions-for-environment-variables


alan20854:
last commit broken ^

PR #54, opened 2021-03-11
[Working] works on test ent w/ ssl ignore

alan20854:


alan20854:
look at previous commits to see how we skip ssl on our test instance

PR #56, opened 2021-04-26
Upgrade Docker Images

alan20854:
Per our Harvard customer, who uses [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) on their Github Enterprise (server) instance, they're currently getting errors running our action. The action was previously working on their original runner, but since centralizing their runners, it's no longer working (erroring on our exec.Command git commands with `exit status 128`).

He spun up two self-hosted runners to test (one with Ubuntu 20.04 LTS vs one with Ubuntu 18.04 LTS), and version 20 works while version 18 does not. Could this be because our docker images are behind in version? What are some potential reasons why we're seeing this behavior? Could version 20 address some compatibility issues that 18 did not have?

If bumping alpine to latest doesn't work, should we just have the image for our binary be ubuntu:20? I forget how long it took to build an ubuntu image, but I think that was the reasoning behind using alpine instead initially.


alan20854:
Also, the customer is happy to re-run our action, so we could tweak which images we're using to get a good balance between what works and build time

louis-nightfall (approved):

Dockerfile
```diff
@@ -15,7 +15,7 @@ COPY . .
 
 RUN CGO_ENABLED=1 GOOS=linux go build -a -installsuffix cgo -o nightfall_code_scanner ./cmd/nightfalldlp/
 
-FROM alpine:3.8
+FROM alpine:latest
```
bradleymcallister97: I am wondering if we should use `latest` as the tag. If we use this then we will always be grabbing the newest alpine image. So we can eventually start pulling images that have a major version increase without having tested it. Not sure if this is desired? 

At one point I thought you could use the tag `3.x.x` or `3.x` and this would always get the latest version on major version `3`. But cannot seem to get this to work locally when pulling the alpine image.

alan20854:
Bug Description + Resolution documented here: https://shorelineai.slack.com/archives/C01BEMYMU2V/p1619558051121100

The issue was the self-hosted runner git version was < 2.18, so the result from the actions/checkout step wasn't what we expect when the version >= 2.18

PR #57, opened 2021-09-16
Use new nightfall API client

bradleymcallister97:


Dhertz (approved):

internal/nightfallconfig/nightfall_config_test.go
```diff
@@ -48,6 +63,8 @@ func TestGetNightfallConfig(t *testing.T) {
 		FileExclusionList:  []string{".nightfalldlp/config.json"},
 	}
 	actualConfig, err := nightfallconfig.GetNightfallConfigFile(workspacePath, testFileName, nil)
+	val, _ := json.Marshal(expectedConfig)
+	fmt.Println(string(val))
```
alan20854: rm print

alan20854 (approved):

internal/nightfallconfig/nightfall_config_test.go
bradleymcallister97: good catch

PR #68, opened 2022-01-05
Support for defaultRedactionConfig

evanfuller:
Add support for `defaultRedactionConfig` to config file

### DIFF
--- a/internal/clients/nightfall/nightfall.go
+++ b/internal/clients/nightfall/nightfall.go
@@ -20,34 +20,30 @@ import (
 )
 
 const (
-	// max request size is 500KB, so set max to 490Kb for buffer room
-	// maxSizeBytes = 490000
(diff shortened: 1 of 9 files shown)


internal/clients/nightfall/nightfall.go
```diff
@@ -87,7 +84,7 @@ func blurContent(content string) string {
 }
 
 func getCommentMsg(finding *nf.Finding) string {
-	if finding.Finding == "" || finding.Detector.DisplayName == "" {
+	if (finding.Finding == "" && finding.RedactedFinding == "") || finding.Detector.DisplayName == "" {
```
bradleymcallister97: If we are redacting the finding, then do we still need to blur the content with the custom `blurContent` function?

Dhertz (approved):

PR #72, opened 2022-06-07
add ability to change annotation level

### LINKED ISSUE #71: Allow annotation level to be configured for Github Action
Currently, findings are always reported in the github action as `failure` level findings. This is hardcoded here: https://github.com/nightfallai/nightfall_code_scanner/blob/92a6b9ab2e9e93a83cb5a6d8179d370813d1a64e/internal/clients/diffreviewer/github/github_service.go#L37

We would like to pilot this action running on our PRs, with the findings reported at a `notice` or `warning` level. This will make the tool slightly less noisy while we evaluate the false positive rate internally and fine tune our configuration.

It would be great if you supported configuring the level that findings should be reported for your github action. From our perspective it would be useful either in the github workflow configuration itself:

```
      - name: Checkout Repo Action
        uses: actions/checkout@v2

      - name: nightfallDLP action step
        uses: nightfallai/nightfall_dlp_action@v2.0.2
        env:
          NIGHTFALL_API_KEY: ${{ secrets.NIGHTFALL_API_KEY }}
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          EVENT_BEFORE: ${{ github.event.before }}
          ANNOTATION_LEVEL: warning
```

or secondarily as a separate entry in the `.nightfalldlp/config.json`. If in the nightfall config, then it might be a little strange unless this is promoted to a full nightfall level concept. The advantage would be it could be configured at a per detection rule level, but this greatly increases the scope of this ticket.

For reference, these are the supported levels ([docs](https://docs.github.com/en/rest/reference/checks)): 
Screen Shot 2022-04-15 at 10 47 59 AM

Dhertz:
fixes #71

### DIFF
--- a/internal/nightfallconfig/nightfall_config.go
+++ b/internal/nightfallconfig/nightfall_config.go
@@ -22,7 +22,8 @@ const (
 	defaultConditionsInfoMessage = "Using default Detection Rule with detectors API_KEY and CRYPTOGRAPHIC_KEY"
 )
 
-var annotationLevelFailure = "failure"
+// AnnotationLevelFailure describes the failure severity to render comments on code
+var AnnotationLevelFailure = "failure"
 var annotationLevels = map[string]struct{}{"notice": {}, "warning": {}, "failure": {}}
 var defaultNightfallConfig = &ConfigFile{
 	DetectionRules: []nf.DetectionRule{
@@ -61,7 +62,7 @@ var defaultNightfallConfig = &ConfigFile{
 			NumCharsToLeaveUnmasked: 2,
 		},
 	},
-	AnnotationLevel: annotationLevelFailure,
+	AnnotationLevel: AnnotationLevelFailure,
 }
 
 // ConfigFile is the struct of the JSON nightfall config file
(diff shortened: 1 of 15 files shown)


martinmr (approved):

internal/nightfallconfig/nightfall_config.go
```diff
@@ -113,5 +118,9 @@ func GetNightfallConfigFile(workspacePath, fileName string, logger logger.Logger
 		nightfallConfig.MaxNumberRoutines = MaxConcurrentRoutinesCap
 	}
 	nightfallConfig.FileExclusionList = append(nightfallConfig.FileExclusionList, nightfallConfigFilename)
+	// must be one of notice, warning, or failure
+	if _, ok := annotationLevels[nightfallConfig.AnnotationLevel]; !ok {
+		nightfallConfig.AnnotationLevel = annotationLevelFailure
```
bradleymcallister97: It would be good if we could have a warning log about this so they can see if they accidentally input an invalid value, but not sure thats easily done here without returning an error which we probably dont want to do.

bradleymcallister97 (approved):

internal/nightfallconfig/nightfall_config.go
```diff
@@ -113,5 +119,12 @@ func GetNightfallConfigFile(workspacePath, fileName string, logger logger.Logger
 		nightfallConfig.MaxNumberRoutines = MaxConcurrentRoutinesCap
 	}
 	nightfallConfig.FileExclusionList = append(nightfallConfig.FileExclusionList, nightfallConfigFilename)
+	// must be one of notice, warning, or failure
+	if _, ok := annotationLevels[nightfallConfig.AnnotationLevel]; !ok {
+		nightfallConfig.AnnotationLevel = AnnotationLevelFailure
+		if nightfallConfig.AnnotationLevel != "" {
```
bradleymcallister97, line 130: Because we are setting `nightfallConfig.AnnotationLevel = AnnotationLevelFailure` wont `nightfallConfig.AnnotationLevel != ""` always be `true`? 

In the log message below we will be logging the `AnnotationLevelFailure` instead of the actual input annotation level

bradleymcallister97 (approved):

PR #75, opened 2023-08-29
Abdeali/intm 2073 support combo detectors in nightfall code scanner

abdealiaden:


internal/clients/nightfall/nightfall.go
```diff
@@ -338,6 +400,34 @@ func (n *Client) ReviewDiff(ctx context.Context, logger logger.Logger, fileDiffs
...
 }
 
+func getFileToScan(fd *diffreviewer.FileDiff) (*fileToScan, error) {
+	fts := &fileToScan{
+		FilePath:         fd.PathNew,
+		ContentToLineMap: datastructs.NewRangeMap(),
+	}
+
+	bufferString := bytes.NewBufferString("")
+	startByteRange, endByteRange := 0, -1
```
bradleymcallister97: I believe these values are technically the codePoint ranges not byte ranges, since we are using the number of chars not bytes.

internal/clients/nightfall/nightfall.go
```diff
@@ -250,45 +286,71 @@ func (n *Client) scanContent(
...
+				break
+			}
+			currentSize += size
+			endIndex++
+		}
+		requestBatch = append(requestBatch, cts[startIndex:endIndex]...)
+		requestBatches = append(requestBatches, requestBatch)
+	}
 
+	// Integer round up division
```
bradleymcallister97: I think we can remove this comment since it no longer applies

internal/clients/nightfall/nightfall.go
```diff
@@ -250,45 +286,71 @@ func (n *Client) scanContent(
...
 	ctx context.Context,
 	logger logger.Logger,
-	cts []*contentToScan,
+	cts []*fileToScan,
 	commentCh chan<- []*diffreviewer.Comment,
 ) {
 	defer close(commentCh)
 	blockingCh := make(chan struct{}, n.MaxNumberRoutines)
 	var wg sync.WaitGroup
+	requestBatches := make([][]*fileToScan, 0, 1)
```
bradleymcallister97: why do we need an array that is only of size 1? Can we not just use `requestBatch := make([]*fileToScan, 0)`?

Also how are we guaranteed that we will only have 1 array of content to scan?

internal/clients/nightfall/nightfall.go
```diff
@@ -250,45 +286,71 @@ func (n *Client) scanContent(
...
+		startIndex := endIndex
+		currentSize := 0
+		for {
+			if endIndex >= len(cts) {
+				break
+			}
+			size := len(cts[endIndex].Content)
+			if size > maxAPIRequestSize {
+				// file diff size is greater than supported by API platform, should not come here
+				logger.Error("terminating the scan early as large file scanning is not supported by API platform")
```
bradleymcallister97, line 316: I believe the file size is more of a Github Action limitation our API should be able to scan large content

internal/clients/nightfall/nightfall.go
```diff
@@ -303,26 +365,26 @@ func (n *Client) Scan(ctx context.Context, items []string) (*nf.ScanTextResponse
 // contains sensitive data
 func (n *Client) ReviewDiff(ctx context.Context, logger logger.Logger, fileDiffs []*diffreviewer.FileDiff) ([]*diffreviewer.Comment, error) {
 	fileDiffs = filterFileDiffs(fileDiffs, n.FileInclusionList, n.FileExclusionList, logger)
-	contentToScanList := make([]*contentToScan, 0, len(fileDiffs))
-	// Chunk fileDiffs content and store chunk and its metadata
+	//contentToScanList := make([]*contentToScan, 0, len(fileDiffs))
```
bradleymcallister97: remove this?

internal/clients/datastructs/rangemap.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+type pair struct {
+	left  int
+	right int
+}
+
+type RangeMap struct {
+	ranges []pair
+	values []int
```
bradleymcallister97: can we combine these 2 arrays into 1 array of a struct? If we add `value int` to the `pair` type then it will represent the same data right? I believe this would optimize when we are doing any array element changing here https://github.com/nightfallai/nightfall_code_scanner/pull/75/files#diff-f0748cd1efda32b04e12ec22884ea129532b00ebdb3cc2334701cdec24e15fc6R50-R53

internal/clients/datastructs/rangemap.go
```diff
@@ -0,0 +1,62 @@
...
+		return errors.New("range should not overlap")
+	}
+	toBeInsertedAt := toBeInsertedAtRight
+	if toBeInsertedAt >= len(r.ranges) {
+		r.ranges = append(r.ranges, pair{left: left, right: right})
+		r.values = append(r.values, value)
+		return nil
+	}
+
+	r.ranges = append(r.ranges[:toBeInsertedAt+1], r.ranges[toBeInsertedAt:]...)
```
bradleymcallister97: can we add a comment above this saying this will insert a new element at index `toBeInsertedAt`

internal/clients/nightfall/nightfall.go
abdealiaden: 1 is just the initial capacity, The Idea here is that in most cases it would be 1, and if needed, Golang's append function will increment the capacity.

bradleymcallister97: oh this is new to me I thought that if you exceed a slices capacity then you would get a panic

internal/clients/nightfall/nightfall.go
bradleymcallister97: Also dont we already check this before calling this func?

internal/clients/nightfall/nightfall.go
bradleymcallister97: I kinda feel like setting the capacity to `1` is confusing, since there are cases where we might exceed that value

internal/clients/nightfall/nightfall.go
```diff
@@ -250,45 +286,71 @@ func (n *Client) scanContent(
...
 	blockingCh := make(chan struct{}, n.MaxNumberRoutines)
 	var wg sync.WaitGroup
+	requestBatches := make([][]*fileToScan, 0, 1)
+	endIndex := 0
+	for endIndex < len(cts) {
+		requestBatch := make([]*fileToScan, 0)
+		startIndex := endIndex
+		currentSize := 0
+		for {
+			if endIndex >= len(cts) {
```
bradleymcallister97, line 310: Feel like we might want to separate this into 2 different cases. 
`if endIndex == len(cts) {break}`
`if endIndex > len(cts) {//something bad happend}`

internal/clients/nightfall/nightfall.go
```diff
@@ -175,6 +183,34 @@ func sliceListBySize(index, numItemsForMaxSize int, contentToScanList []*content
...
 
+func createCommentsFromScanRespForFiles(inputContent []*fileToScan, resp *nf.ScanTextResponse, tokenExclusionList []string) []*diffreviewer.Comment {
+	comments := make([]*diffreviewer.Comment, 0)
+	for j, findingList := range resp.Findings {
+		for _, finding := range findingList {
+			if finding.Finding != "" && !isFindingInTokenExclusionList(finding.Finding, tokenExclusionList) {
+				// Found sensitive info
+				// Create comment if fragment is not in exclusion set
+				correspondingContent := inputContent[j]
+				exists, lineNumber, _ := correspondingContent.ContentToLineMap.Find(int(finding.Location.ByteRange.Start))
```
bradleymcallister97: I think here we should use `finding.Location.CodePointRange.Start` as everywhere else we are counting the number of chars not bytes.

internal/clients/nightfall/nightfall.go
```diff
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/nightfallai/nightfall_code_scanner/internal/clients/datastructs"
```
bradleymcallister97: import should be below with the other github imports

internal/clients/nightfall/nightfall.go
```diff
@@ -303,26 +365,26 @@ func (n *Client) Scan(ctx context.Context, items []string) (*nf.ScanTextResponse
...
-					logger.Error("Error chunking git diff")
-					return nil, err
-				}
-				contentToScanList = append(contentToScanList, chunkedContent...)
-			}
+		file, err := getFileToScan(fd)
+		if err != nil {
+			return nil, err
 		}
+		if len(file.Content) > maxAPIRequestSize {
```
bradleymcallister97, line 374: this seems like a new constraint that did not exist before these changes. Should we just chunk the file?

bradleymcallister97 (approved):

internal/clients/nightfall/nightfall.go
abdealiaden: yes, we check this early on. This is just a failsafe

internal/clients/nightfall/nightfall.go
abdealiaden: It will never be >len(cts), as endIndex is incremented by 1 at a time in this loop.

internal/clients/nightfall/nightfall.go
abdealiaden: Chunking the file would potentially miss the findings from combo detectors/multiline detectors. 
But, I agree that it is better to scan by chunking the file, rather than just skipping it altogether. 
I have discussed this with product team (Mayank), and we have agreed that the files (diff) which is greater than 512kb, can be skipped altogether with the appropriate logs, as it would be very rare for nonbinary file to have the limit and the plan is to create a github action which communicates directly with github dlp service in the long term.

internal/clients/nightfall/nightfall.go
abdealiaden: Ok, i have reverted to 0 capacity

